PR #20, opened 2025-01-15
Add calendar plugin

jason-s13r:
I made a handful of changes since the images I put on the discussion thread.

- Each ICS calendar feed has their own colour. Indicated by a dot.
- black text on the white background is easier to read, so all day events are distinguished with coloured border rounded rectangle with no fill.
- row at top for (render) time + month-year.
- Today has time of day indicator with sunrise/noon/sunset points highlighted. And Now is yellow circle (sun) between sunrise-sunset, or white crescent (moon).
- Settings uses Leaflet with Open Street Maps to pick a location. This is used for Timezone and Sun time calculation.
- Settings page stores form fields in localStorage.

image

Things I would like (cbf):
- `python-weather` to show weather report in the top row between time & month.
- some cleanup in image_calendar.py; splitting up render func, the sun time indicator line, etc.
- plugin icon doesn't look nice next to the others.


SillyHats:
I would like to request an option to disable the diagonal lines through past days, if possible.

jason-s13r:
image

:)

SillyHats:
One more request regarding the to be added weather report in the top bar using python-weather.

Include the `feels_like` reading in addition to `temperature` or make it a toggleable option. 
https://python-weather.readthedocs.io/en/stable/forecast/weather.html

Also, the ability for fahrenheit would be much appreciated.

Thanks for creating this plugin!

jason-s13r:
Maybe. Haven't done anything on the weather idea yet.

I should probably also consider giving the option to or using system locale to determine which units to use.  Obviously I built this with my preference for 24 hour clock time and Monday as start day.

SillyHats:
Thanks for the response.

Using system locale to determine preferences would simplify things. 

However, as someone living in imperial unit land who prefers a Monday start to the calendar week, but also 12-hour clocks and °F additional settings to accommodate idiosyncratic preferences would also be nice.

danielboos1:
Not sure if you would be interested to updating the plugin but I have designed a possibility nicer way of displaying data. Given that the display sizes are fixed you can use an image and place the text over it no need for it to be responsive really but even then, it's just a bunch of text and rectangles.
Untitled-1

jason-s13r:
@danielboos1 looks nice.

I was replicating my paper printed fridge calendar where we write events on the day squares. And wanted to retain that visibility of the events each day.

Though because of the size of the screen, readability is compromised by text size, I having been thinking of a view that shows the simple/numbers only calendar (ie TextCalendar) in a corner.
An area for today's agenda.
An area for up to the next 7 days.
This also means not wasting space for the past.

I like the idea of the event count indicated in the simple calendar.
The weather also looks nice.

fatihak:
Hey, thanks for building out the calendar plugin—it looks really good! I haven’t had a chance to test it yet as I’ve been under the weather and traveling recently, but I’m planning to review it in the coming week.

I also really like all of the suggestions but just a heads-up: I want to eventually introduce modular layouts for the project, which will allow multiple plugins (e.g., weather and calendar) to be displayed simultaneously. So i wouldn't recommend adding more information/features to the calendar plugin itself, but I’m not sure when modular layouts will be ready so it’s entirely up to you.

Including an image here on what i mean by modular plugins:
modular_plugins

jason-s13r:
@fatihak cool, I'll skip weather and let this plugin stay focused on keeping it calendar & events only.


Could be worth avoiding needing to edit plugins.json and the requirements*.txt files somehow.
Maybe each plugin should have a manifest file in its folder?

fatihak:
The idea behind having a single, central plugins.json file was to create a user-specific configuration similar to config.json. This would eventually allow enabling/disabling individual plugins and storing plugin-specific data but I'm open to any ideas or suggestions for this.
I think it's a good idea to have a requirements.txt file per plugin, we would need to have the install script navigate through the plugins directory and aggregate all the dependencies. I was going to ask you if you had any suggestions on how to best handle the new requirements being added in this PR and also for future changes. Currently, if users people do a git pull, the calendar plugin would fail to initialize unless they re-install the pip dependencies.

SillyHats:
I'm a big fan of the paper calendar replacement look as that is how I usually use the online calendar personally and for sharing among family. This is a great way to display that information without needing to use a web browser.

The plan for modular plugins seems great. In this case the minimal weather info that could be tucked into the Month/Time  line at the top of the display would be a much more efficient use of space than an entire quarter of the display.

In keeping with the paper calendar replacement look while maximizing the future looking information sliding the current week to the top row would always shown 4 or five weeks in the future.

After getting this branch installed I got this error if I selected a location on the map and the coordinates populated below the map and then I updated the display followed by entering calendar URL(s) without selecting on the map again before clicking update now.
Error! An error occurred: could not convert string to float: ''

Installation process for future reference
`git clone -b add-calendar-plugin https://github.com/jason-s13r/InkyPi.git`

https://github.com/fatihak/InkyPi/blob/main/docs/troubleshooting.md
```
source "/usr/local/inkypi/venv_inkypi/bin/activate"
cd InkyPi
pip install -r install/requirements.txt
deactivate
sudo systemctl restart inkypi.service
```

jason-s13r:
@SillyHats 
>Error! An error occurred: could not convert string to float


Perhaps the latitude or longitude is empty string in Python. Though I think before I added float() around the settings.get() the TimezoneFinder threw that.

SillyHats:
One more potential bug report. 

When adding a public calendar URL published in UTC+0 the InkyPi seems to display the time UTC without adjusting to the local offset. For comparison google calendar converts to the local time zone when adding this link as an external calendar.

Here is the calendar link if you want to see it in action https://www.philadelphiaeagles.com/api/addToCalendar/ag/s?text=

fatihak:
@jason-s13r I was able to test the plugin and it works really nicely! Some issues I've noticed:
1. I realized that doing a git pull post-install with these changes will not import/display the new plugins introduced since InkyPi is looking at the user-specifc plugin file in src/config/plugins.json (this is my bad on my part).
2. As @SillyHats pointed out, the calendar is displaying the time in UTC .I was considering moving the timezone setting from the clock plugin to the device settings page so its accessible by all plugins rather than having the setting in each plugin that needs it, might be a good time to do that so the calendar can use it.
3. This will require a pip install for new plugins introduced, not sure what the best way to streamline this for anyone pulling the changes, maybe introduce an update.sh script? 
4. There are some minor UI improvements that can be made but its not that big of a deal, like resizing the location selection map or making it visible only when selected.
I can assist with addressing some of these, let me know what your thoughts are. We can also discuss further on discord or another platform.

jason-s13r:
Sure, discord, email, telegram, etc.

I'll close this PR for now and I'll get that time display fixed up.

fatihak:
Can you add me on discord, my username is fmetinak. I started on some of the changes on my side, we can discuss there and coordinate.

Porco-Rosso:
Just thought I would leave a comment here:
I found public ical links to be a hassle, both from the security perspective, and from the need to share multiple calendars to get it working.
Instead I opted for sending the json data directly from iCalPal. Might help in some way.
https://github.com/Porco-Rosso/eCal

AlessandroBernini:
I’m surely late for this but, couldn’t you guys take into account the idea to take the branch back to development? With a link to google calendar it could be really useful and it honestly looks good and modern.

PR #33, opened 2025-02-08
Added Countdown plugin

stuartw0:


src/config/device.json
fatihak, line 1: This file shouldn't be committed

src/plugins/countdown/Thumbs.db
fatihak, line 1: This file shouldn't be committed

src/static/images/current_image.png
fatihak, line 1: This file shouldn't be committed

src/plugins/countdown/countdown.py
fatihak, line 1: A lot of the logic for generating images, adding frames, and wrapping lines here is duplicated from the AI Text plugin. To avoid maintaining the same code in multiple places, I think it would be beneficial to move these functions to a utility file/class. Long-term, I think we need to introduce a shared "image options" section in the UI can be configured for any plugin.
If you can tackle those changes feel free to do so, otherwise I would suggest removing them here.

fatihak (commented):
Thanks for the new plugin and PR! I left a few comments, but I’m planning to hold off on merging any pull requests until the Playlists feature is completed and merged. That update introduces the ability to edit existing scheduled plugins, which will require additional logic to populate the HTML form with existing settings.

src/plugins/countdown/countdown.py
stuartw0: Makes sense to remove the duplicate code.

I'll close this PR for now and revisit once the playlist feature is implemented. Definitely think this need to include a "play at x" in addition to "play every x" scheduling.

stuartw0:
Makes sense to remove the duplicate code.

I'll close this PR for now and revisit once the playlist feature is implemented. Definitely think a "play at x" in addition to "play every x" scheduling is required.

PR #39, opened 2025-02-14
Playlists

fatihak:
### Playlists  
- Playlists can now be viewed and edited on the new Playlists page, accessible via the icon in the top right of the home page.  
- You can create playlists with a Start Time and End Time—plugins added to the playlist will cycle during this interval.  
- Overlapping Playlists:
  - If multiple playlists overlap, the one with the shortest time interval (end_time - start_time) takes priority. This allows for a default playlist throughout the day while defining more specific time periods for certain plugins.  
### Plugin Instances  
- The "Schedule Refresh" button is now "Add to Playlist", which creates a Plugin Instance.  
- Plugin instances have:  
  - Playlist Name – The playlist to add to.  
  - Instance Name – A unique name for this instance.  
  - Refresh Settings:  defines how often the data/image for this plugin should be refreshed/generated
    - "Every X minutes/hours" – Periodic refresh on an interval.  
    - "Daily at TIME" – Ideal for less frequent updates (e.g., Newspaper, AI-generated content).  
  - Note: Plugins only refresh when they are about to be displayed

### Display Refresh Improvements  
- Skipping unnecessary refreshes: The code now checks if a new image is the same as the currently displayed one by comparing their hashes and skips the refresh if they are the same.  
- Empty time slots: You can create playlists without any plugins, or set time periods without a playlist if you want to avoid updates during specific times, such as overnight.

### Image Upload Plugin  
- The Image Upload plugin now supports uploading multiple images. When added to a playlist, the images cycle on each refresh if multiple are provided.  

## Important Device Settings  
- Time Zone: Ensure the time zone is set correctly, as this is to compare with the playlist start/end times.  
- Plugin Cycle Interval: Defines how long each plugin is displayed before cycling. Manual updates will also respect this duration. This defaults to 1 hour.

## Testing Instructions  
- Create as many playlists with different time periods & add plugins to them.  
- Monitor your display and the  "Playlists" page to verify:  
  - The correct playlist is active. 
  - Plugins are displayed as expected. 
- Edit/delete playlists and plugin instances, ensuring changes are reflected.  
- Test multi-image uploads

To pull the changes on your PI:  
```bash
cd InkyPi
git pull
git checkout Playlists
sudo systemctl restart inkypi.service
```

Let me know if you encounter any issues or have any suggestions! Feel free to leave comments on this PR if you try it out so I know everything is working smoothly. I plan to merge the changes in the next 2-3 days.


boessi:
Just started playing around with it for a bit. Just some stuff the I found so far. Loving the project and this new Playlist feature so far.

- Uploaded multiple images to the Default Playlist with a Interval of 5min, no update happeing ~10min later (1 Hour plugin cycle). When creating multiple instances (with multiple images) and a lower plugin cycle it will switch and rotate (Testing with 1 Min plugin cycle / 5 min refresh interval).
- Refresh Interval on a Instance can not be adjusted after created
- Would be nice to move a Instance from one Playlist to another
- Would be nice to define a random selection of a Image

Will keep it running and try to add different Test Cases

nov1n:
Great feature!

Some feedback:
- When adding a plugin to a playlist with a name that already exists the error message is "Failed to add plugin", which could be more descriptive, like: "Plugin with title 'foo bar' already exists in playlist "default".
- It would be nice if I could trigger a plugin to update now from the "edit playlists" page for testing purposes.

Will continue to test and update the thread with any new feedback.

fatihak:
@boessi Thanks for testing it out and for the feedback!

For the first point, that behavior is expected due to the plugin cycle interval. Since the 7-color displays have long refresh times, this setting enforces a hard rule for the display update frequency, regardless of the refresh interval set for individual plugins. I realized this might be a little confusing with the plugin refresh setting so I tried adding the ⓘ tooltips to help clarify, but open to any suggestions for renaming them.

I think the other suggestions are great enhancements and had considered them as well, but I dont think I'll be including them in this update. However, I’ll create an issue to track playlist improvements for the future. Let me know if you run into anything else!

fatihak:
@nov1n Thanks for the feedback!
I updated the error message when it fails to create a plugin instance with an existing name.

For your second point, I added a "Display Now" button for the plugin instances in the playlist page that triggers an immediate update. This should be really helpful since it's difficult to tell if a plugin instance is failing to refresh other than monitoring the logs or checking the latest refresh label, so this can be used for testing. Note that the button respects the plugin instance's refresh settings so it won't force a refresh if the instance’s rule hasn't triggered yet.

I also found and fixed a bug related to plugins not refreshing when using a scheduled time refresh. Let me know if you find anything else!

PR #47, opened 2025-02-20
updated to resolve #46

### LINKED ISSUE #46: Inconsistency between error check and error wording
When  adding to playlist and "_Instance Name_" contains a number, the error:

`Error! Instance name can only contain alphanumeric characters and spaces`

is shown.

The code that is checking this in playlist.py only allows alphabet letters (`char.isalpha()` ) or spaces (`char.isspace()`). Numeric characters ( `char.isnumeric()` ) is not allowed

```

if not all(char.isalpha() or char.isspace() for char in instance_name):
            return jsonify({"error": "Instance name can only contain alphanumeric characters and spaces"}), 400
```

Either add `char.isnumeric()` or change the error wording.

stuartw0:
Update to resolve #46

### DIFF
--- a/src/blueprints/playlist.py
+++ b/src/blueprints/playlist.py
@@ -27,7 +27,7 @@ def add_plugin():
             return jsonify({"error": "Playlist name is required"}), 400
         if not instance_name or not instance_name.strip():
             return jsonify({"error": "Instance name is required"}), 400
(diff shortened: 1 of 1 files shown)


fatihak (approved):

fatihak:
Looks good, merged

PR #52, opened 2025-03-11
Add xkcd plugin

nov1n:
The plugin displays a random comic from the [xkcd](https://xkcd.com/) api.

No idea if you're interested in adding this to inkypi core, but I've been having fun with it.

The page on plugin development was very clear, thanks for adding that. 👍


fatihak:
Thanks for the PR! I've been wanting to add a comic plugin and this is really good! I’d like to add options for different sources for comics since I'm guessing people will ask for others. Also noticed the dimensions are hardcoded here, which won't work with other display sizes or orientations.

If you want to tackle those changes go ahead, otherwise I might use this as a base to make implement it sometime soon.

nov1n:
I don't really have time to work on this plugin, but feel free to use it as a starting point / inspiration!

fatihak:
Merged in https://github.com/fatihak/InkyPi/pull/256 that adds a Daily comic plugin with 6 different comic options. It only displays the latest comic from their rss feeds right now, but I want to eventually add the ability to get random comics but I think this is a good start. Thanks for the inspiration! Closing this PR.

PR #59, opened 2025-03-16
Added Options for Inverted Orientations

vifair22:
Simple update;

 - Added the ability to select an inverted version for each of the orientations allowing users to select the original rotation values of 0, 90 and now 180, 270
 
 I'm open to changing the display name or the setting value. Might be worth changing the settings values to degrees of rotation to make localization easier in the future.
image


fatihak:
Can you add this as a separate setting, maybe just "Inverted" with a checkbox next to the orientation setting?

fatihak:
Closing this in favor of https://github.com/fatihak/InkyPi/pull/118, which has the change requested

PR #99, opened 2025-04-19
Add moonphase to weather plugin

Ch3my:
Added Option to show moon phase in forecast, included in plugin settings can deactivate. In forecast card displays icon and percentage of moon read from the same API 

This is an example image

current_image


fatihak:
This looks really nice! While testing I noticed the moon phases are displayed even if I don't check the box in the settings

Ch3my:
You're right, I fixed it

fatihak:
Merged! Thanks for the PR

PR #101, opened 2025-04-21
Make plugins more plug-in

doofmars:
Removes the need for a central plugins.json
Each plugin has a `plugin-info.json` which will be loaded on startup

With this plugins can be a separate git repos and installed via git by running this command in the inkypi repo root 
`git clone https://my-awesome-plugin/a-plugin.git plugins/a-plugin`

For now I decided to keep it simple and close to your current implementation.
Still this change will improve plugin development.


preshit:
Oh this is very clever. Would love to see this in the main repo.

PR #118, opened 2025-05-09
Add option to invert images

CattoSalad:
I saw [Added Options for Inverted Orientations #59](https://github.com/fatihak/InkyPi/pull/59) and decided to implement the changes you requested.

It's now a checkbox next to the Orientation as per requested.
image

This feature gives users the flexibility to use any frame or mounting position without worrying about the display orientation.

I went with `inverted_image` for the setting, I felt like it was the most descriptive option. I also updated `change_orientation` to support this, and made sure it defaults to `false`.
I tested it and it works on my end, but do feel free to test it yourself and share your thoughts. 

One thing to note: the preview remains upright, which makes sense in my opinion, but let me know what you think.

What's left: Adding this to the plugins and the plugin template, which I am happy to do after some feedback.

### DIFF
--- a/install/config_base/device.json
+++ b/install/config_base/device.json
@@ -1,6 +1,7 @@
 {
     "name": "InkyPi",
     "orientation": "horizontal",
+    "inverted_image": false,
     "scheduler_sleep_time": 60,
     "startup": true
 }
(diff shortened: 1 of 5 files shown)


fatihak:
Looks good, was able to test it out and worked as expected!

I think the preview should remain upright, and don't think there are any changes needed to the plugins since the changing of orientation/inverting takes place after the plugins generate an image. I tested with the Clock, Newspaper, and Weather plugins and they all worked. Thanks for the PR!

PR #120, opened 2025-05-11
Add option for manual location name entry

bluenadas:
image
current_image

This feature gives users the ability to manually assign a city and State/Country name, if desired. For my specific location, I get a rather generic name when using the OpenWeather API location based on my lat/long (county name, state). 

Having the ability to manually enter a location makes the presentation better.

Code is written to bypass OpenWeather location API call if manually entry is desired.

### DIFF
--- a/src/plugins/weather/settings.html
+++ b/src/plugins/weather/settings.html
@@ -15,6 +15,22 @@
 </div>
 
 <div class="form-group">
+    <label class="form-label">Manual Location Name: </label>
+    <div class="form-group">
+        <input type="checkbox" id="enableName" name="enableName" onclick="this.value=this.checked ? 'true' : 'false';">
+        <span>Enable?</span>
+    </div>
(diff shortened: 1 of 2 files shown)


bluenadas:
Moved to draft as I notice the location data is not persistent in the Playlist.

fatihak:
Thanks for the PR! I get the idea behind displaying a custom location name, but I feel it adds extra settings for something that's mostly cosmetic. The current setup pulls the correct location automatically, so I’d like to keep it simple and stick with that.

cello8619:
@fatihak, can you revisit this?  I think a lot of people would like this, for example, it is very common in Florida to have an address in a specific city; however, technically, the location is an unincorporated county.  So the display shows the county name and not the location name, which makes sense.  In my example, it shows Palm Beach County, Florida, when it would be much better to show Boca Raton, Florida.

PR #127, opened 2025-05-18
Weather: add option for last refresh time

cookandy:
Adds an option to display last refresh time in the upper right of the weather plugin. 

(My linter also decided to fix a few formatting issues.)

### Example

current_image (1)

### DIFF
--- a/src/plugins/weather/render/weather.css
+++ b/src/plugins/weather/render/weather.css
@@ -25,6 +25,20 @@
   container-type: size;
 }
 
+.last-refresh {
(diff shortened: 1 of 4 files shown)


Reid-Kornman:
@cookandy - Any chance you could remove the date from the "Last refresh" label as part of this update? It doesn't play well when you have a longer town name and is already known by the date shown below the town name.

image

cookandy:
@Reid-Kornman I'll try to find some time this week. Perhaps we can make the date an optional value.

PR #128, opened 2025-05-23
Adding Option for Min/Max for current day and Option to show last update

vendolis:
This implements two new options:

- Current Day Min/Max (shows above the current temperature)
- Last Update (shows behind the current data)

Both are implemented as option in the settings.

weatherSettings
weatherImage


sostendorf:
Nice job with this.  I was just about to do something similar so I'm glad I wasn't the only one annoyed that there was no high/low for the current date... ;)

fatihak:
Thanks for the PR, I think its nice to be able to see the min/max for the current day. I merged an earlier PR that added the latest refresh date/time on the top right. Can you revert the changes to only have the min/max changes? Also visually I would prefer the min/max to be under the "Feels Like".

Reid-Kornman:
@vendolis - do you need any help on this? I'd love to get this update too.

tobiasvogel:
I made the changes requested by @fatihak: moved the min/max below the actual temperature and removed the "last refresh" from this PR. In order not to open another PR for this very issue and to keep things clean, I attached this patch
[weather.patch](https://github.com/user-attachments/files/21351049/weather.patch)
 which works against the latest commit.

fatihak:
Merged in the changes from the patch (with some slight adjustments), feel free to pull the latest changes and test it out. Closing this pr.

PR #148, opened 2025-06-03
Add 12h/24h time format in device configuration option (which is used in weather plugin)

JeremyCottontail:
Adds a new configuration option to allow users to choose between 12-hour and 24-hour time formats for displaying times throughout the weather plugin.

### Changes
- Added time format selection dropdown to settings UI (12h/24h)
- Updated sunrise/sunset times to respect selected format
- Updated hourly forecast times to use selected format  
- Updated last refresh timestamp to use selected format
- Added helper method `format_time()` for consistent time formatting
- Set 12h format as default for backward compatibility

### Screenshots
settings
current_image

### DIFF
--- a/src/plugins/weather/weather.py
+++ b/src/plugins/weather/weather.py
@@ -71,7 +71,7 @@ class Weather(BasePlugin):
         timezone = device_config.get_config("timezone", default="America/New_York")
         time_format = device_config.get_config("time_format", default="12h")
         tz = pytz.timezone(timezone)
-        template_params = self.parse_weather_data(weather_data, aqi_data, location_data, tz, units, settings)
+        template_params = self.parse_weather_data(weather_data, aqi_data, location_data, tz, units, time_format)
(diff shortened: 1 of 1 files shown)


fatihak:
Thanks for the PR! Can the setting be moved to the device settings page since we already have the timezone selection there, and so this can be used across plugins?

JeremyCottontail:
@fatihak I implemented the changes. I moved the setting into the device settings page and refactored the rest.

Screenshot 2025-06-04 094447

JeremyCottontail:
@fatihak I implemented the changes. I moved the setting into the device settings page, removed it from the weather plugin settings page and refactored the rest.

Just as a note. I didn't touch the other plugins.

Screenshot 2025-06-04 094447

fatihak:
That's fine, I believe the only other plugin that displays time is the clock plugin which can be updated later. Looks good, merged

PR #153, opened 2025-06-07
Padding rendering strategy for image_upload (padding the image)

anton-ovcharenko:
Hi, @fatihak 
Add a new strategy for image rendering to the image_upload plugin: fitting a large or unproportional image to the frame size without cropping it.
зображення

Result looks like:
зображення


phattmatt:
I've installed this PR to a new InkyPi instance and tested it with a few different images (different sizes and ratios), and with the screen horizontal and vertical.

The results look good to me, with the images rendered as I expect.

Also checked the behaviour of the plugin when not choosing the new option, and it appears to behave as before.

The new code also behaves well on the upgrade case (i.e. with settings created from before the plugin was updated). When a plug in instance already exists the lack of the setting is handled.

I see no reason not to merge this PR as it takes no feature away, only adds a new option, which users are free to ignore or take advantage of at their leisure.

phattmatt (approved):
I've installed this PR to a new InkyPi instance and tested it with a few different images (different sizes and ratios), and with the screen horizontal and vertical.

The results look good to me, with the images rendered as I expect.

Also checked the behaviour of the plugin when not choosing the new option, and it appears to behave as before.

The new code also behaves well on the upgrade case (i.e. with settings created from before the plugin was updated). When a plug in instance already exists the lack of the setting is handled.

I see no reason not to merge this PR as it takes no feature away, only adds a new option, which users are free to ignore or take advantage of at their leisure.

anton-ovcharenko:
Thank you for your time @phattmatt !

anton-ovcharenko:
@fatihak would you mind reviewing this PR?

fatihak:
Looks good! Thanks for the PR, merged

PR #160, opened 2025-06-13
Increase value for MAX_FORM_PARTS to fix 413 Request Entity Too Large

anton-ovcharenko:
During the saving of a large configuration (specifically for the Image Upload plugin with over 1,400 images), we encountered the following error:

    Error! An error occurred: 413 Request Entity Too Large: The data value transmitted exceeds the capacity limit.

This appears to be caused by the default value of the MAX_FORM_PARTS configuration parameter in Flask, which is currently set to 1_000.

At the same time, journalctl logs the following:

    Jun 10 00:24:52 inkypi inkypi[2780]: 00:24:52 - INFO - werkzeug - 192.168.0.105 - - [10/Jun/2025 00:24:52] "PUT /update_plugin_instance/foto HTTP/1.1" 500 -

Proposed Solution:
To increase the MAX_FORM_PARTS value up to 10_000 to accommodate larger configurations.

### DIFF
--- a/src/plugins/ai_image/settings.html
+++ b/src/plugins/ai_image/settings.html
@@ -56,10 +56,12 @@
             document.getElementById('randomizePrompt').checked = pluginSettings.randomizePrompt || false;
 
             // Populate text model
-            document.getElementById('textModel').value = pluginSettings.textModel;
+            document.getElementById('imageModel').value = pluginSettings.imageModel;
 
             // Populate quality
-            document.getElementById('quality').value = pluginSettings.quality;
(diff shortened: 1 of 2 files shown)


fatihak:
Were you able to test this with your use case and see it working? I'm fine with merging this but not sure if the image upload plugin and the way it handles images will be able to handle that many files.

anton-ovcharenko:
Yes, I tested it on my InkyPi with over 1,400 images, and it worked as expected. The image_upload plugin handled the files without any issues in my use case after the fix.

fatihak (approved):

PR #164, opened 2025-06-16
BBC News plugin

chrisc123:
Hi - thanks for the project!

Not sure if you'll be interested in it or not but I put together a plugin to display BBC News headlines using their various public RSS feeds.

The CSS in particular is a bit of a 'vibe-coded' monstrosity but I have tested (and refined!) how it renders against all the Inky display sizes and it seems to result in quite a nice layout, to my eyes.

I did end up having to get the plugin to encode CSS and fonts statically in the HTML to get my rendering test script to work properly, I don't think there are major negative consequences to this other than a slightly larger HTML file on the disk for a bit?


fatihak:
Hi, thanks so much for the PR! I’ve actually been thinking about building an RSS feed plugin, so it’s great to see your implementation. That said, a BBC News-specific plugin might be a bit too narrow in scope, so I likely won’t merge it in as-is.

fatihak:
Closing this due to the limited scope as explained above, feel free to open a PR if you end up implementing a more generic RSS feed plugin, otherwise I added it to the InkyPi trello board!

PR #168, opened 2025-06-21
Add image folder plugin

RoliMG:
loads images from a folder on the rapberry pi
horizontal and vertival images are supported they should be placed in their corresponding folders
picks an image at random and prevents picking shown pictures again
resizes the image and adds padding with a blurred version of the image


acines:
What has happened with this idea? I was really hopping to be implemented 🙂

RoliMG:
Wasn't really ready yet, but I can open another PR if people need it

acines:
I would really like this functionality, because it opens more options, like mounting network share to the folder.

fatihak:
I merged in https://github.com/fatihak/InkyPi/pull/255 in favor of the changes here, but thanks for the PR!

PR #171, opened 2025-06-23
Improved mobile rendering of the Playlist page

phattmatt:
Added an additional DIV and CSS style to group the action buttons together so they move to a new line when the Playlist item is too long.

Example of how this will look on a narrow device after the change:

Screenshot_20250623-163623

'Normal' width rendering should not be affected, and should look the same.


fatihak:
Looks good!

PR #173, opened 2025-06-23
Wikipedia Picture of the Day Plugin

phattmatt:
Inspired by the NASA Astronomy Picture of the Day plugin I've written a Wikipedia Picture of the Day plugin.

- Date can be selected, or set to be randomly chosen
- Shrink to fit option for images that are too big to fit on the screen 

Wikipedia publish pictures from many subjects, so adds something in addition to the Astronomy option.

I've added a feature that allows you to choose to shrink the images to fit the screen dimensions, which can help fit the whole picture into view. Depending on the image, the screen resolution, and the orientation, this may work great....or not.

Feedback, suggestions, and comments welcome!

Thanks.


fatihak:
This is a great plugin—thanks for the PR! Just a couple of quick notes:
- Toggling the “Shrink the image to fit the display” option seems to enable/disable the date input. I’m guessing that’s unintentional?
- One small suggestion: I think it might be simpler if the plugin just defaults to showing today’s image, with an optional toggle to randomize the date. I know the NASA APOD plugin includes a date input too, but personally I don’t think it's needed here. Up to you though—it works fine either way.

phattmatt:
Thanks for the notes.

> Toggling the “Shrink the image to fit the display” option seems to enable/disable the date input. I’m guessing that’s unintentional?

Lol - Yes, it is unintentional. I will fix.

> One small suggestion: I think it might be simpler if the plugin just defaults to showing today’s image, with an optional toggle to randomize the date. I know the NASA APOD plugin includes a date input too, but personally I don’t think it's needed here. Up to you though—it works fine either way.

Yes, I agree, the default should be a blank date, which means a plugin instance will always get the current day. 

I would still keep the date selection as it gives the option to the user if they want to put a specific date in.

phattmatt:
I've push updates to address both notes.

fatihak:
Looks good! Merged

PR #184, opened 2025-06-30
Fix image brightness setting

ahellbach:


phattmatt (approved):
Looks good. 

I confirmed the Brightness setting was ignored and this change fixes that.

PR #187, opened 2025-07-01
Add a Download Logs feature to Settings Page

phattmatt:
Add a "Download Logs" button to the Settings page.

Logs are read via a python systemd module ('cysystemd') so should support reading from volatile (default) or persistent (survives reboots) storage. The download button currently downloads the last 24 hours worth of log entries, but the supporting function in 'settings.py' accepts a parameter, so this can be changed, or support a feature to allow user selectable range being added later.

Also changed the colours of the 'Reboot' and 'Shutdown' to indicate their relative consequence levels.

image

Comments and suggestions are welcome.

### DIFF
--- a/install/requirements.txt
+++ b/install/requirements.txt
@@ -10,3 +10,4 @@ openai==1.58.1
 numpy==2.2.1
 icalendar==6.3.1
 recurring-ical-events==3.7.0
+cysystemd==2.0.1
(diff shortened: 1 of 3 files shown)


fatihak:
I'm traveling for the next 10 ish days, so I wont be able to test this but the changes look good. Feel free to merge if it works for you.

fatihak (approved):

PR #193, opened 2025-07-03
Add emoji support for Calendar Event Titles

SillyHats:
Added fonts-noto-color-emoji to debian-requirements.txt for emoji support, primarily on the calendar plugin when calendar event titles include emoji
current_image


fatihak (approved):

fatihak:
Was able to test this and see it working, thanks for the simple fix!

PR #195, opened 2025-07-04
[New Plugin] Unsplash

el:
Adds support for the unsplash photo api.

SCR-20250704-obrv

Fun fact: I created this using vibe coding : )


phattmatt:
Hi @el,

I've just tested this plugin locally and it looks pretty good to me. 

I have a few notes:

1. I suspect the '.vscode/settings.json' is not needed, and has been accidently included. Consider removing it from the PR.
2. Naming the '.env' entry "UNSPLASH_SECRET" but actually requiring the "Access Key" could be a little confusing. Consider naming the entry "UNSPLASH_ACCESS_KEY".
3. Having a plugin specific readme is a good idea. Consider also adding notes to the 'docs/api_keys.md' file in line with the other plugins that need API keys, so it makes it easy to find the instructions.

el:
Thanks for the suggestions. Should be done

fatihak:
This is pretty cool, definitely going to be using this plugin! Was able to test and everything works as expected, thanks for the PR!

PR #223, opened 2025-07-19
Added fix to troubleshooting.md for seg fault rpi0w

VRvinny:
Added steps to use wkhtmltopdf instead of the chromium headless mode as a fix for the RPi0W to prevent seg faults from happening


fatihak:
Going to reject this addition since using wkhtmltoimage with the project's current state wouldn't work with the weather and calendar plugins that use the screenshotting feature.

PR #227, opened 2025-07-21
Quick and dirty Logfile viewer plugin

tobiasvogel:
An easy way to view the logfile from the webinterface without downloading the log first as an alternative to the Download button in the settings.


fatihak:
This isn't the intended use for a plugin, as those are meant to update the display. I think it would be better to add this into the settings page and either have a button that loads the logs in another page or a popup window.

PR #243, opened 2025-07-30
Adding a randamization option to the Photo Upload Plugin

zapper1942:
Problem:
I have a playlist with a big photo library and I wanted a way to randomly cycle through them. Existing functionality would iterate through them in upload order but I wanted a way to randomize that display order.
When Randomization is set the initial picture displayed is also randomized

Solution:
Add a new Randomize toggle in the HTML and call it in the image upload plugin.

Example:
image

### DIFF
--- a/src/plugins/image_upload/image_upload.py
+++ b/src/plugins/image_upload/image_upload.py
@@ -1,4 +1,4 @@
-from plugins.base_plugin.base_plugin import BasePlugin
+from ...plugins.base_plugin.base_plugin import BasePlugin
 from PIL import Image, ImageOps, ImageColor
 from io import BytesIO
 import logging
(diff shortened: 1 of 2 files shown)


src/plugins/image_upload/image_upload.py
```diff
@@ -22,8 +17,29 @@ def generate_image(self, settings, device_config):
...
+        # Get the current index from the device json
+        img_index = settings.get("image_index", 0)
+        image_locations = settings.get("imageFiles[]")
+
+        if img_index >= len(image_locations):
+            # Prevent Index out of range issues when file list has changed
+            img_index = 0
+
+        if settings.get('randomize') == "true":
+            img_index = random.randint(0, len(image_locations))
```
fatihak, line 34: I believe this should be len(image_locations) -1 to prevent list index out of range error since random.randint is inclusive of the range.

zapper1942: Yup, found this out when testing my folder plugin, fixed

fatihak (approved):
Looks good! Thanks for the PR

Issue #245, opened 2025-08-02
Moon phases disappear from weather plugin


ian009:
The moon phase data disappears from the display when the playlist weather plugin is modified 

The moon phase should not disappear if it is still selected

Configure the weather plugin with moon phase selected
Add it to the playlist & run
Confirm that moon phase is displayed
Modify any parameter C to F or how may days to show forecast
Do not uncheck the Moon phase
After saving & when the display refreshes the moon phases are not displayed

Please see screenshots

<img width="399" height="232" alt="Image" src="https://github.com/user-attachments/assets/80d608a9-c971-496e-b012-67e532ff779e" />
<img width="400" height="230" alt="Image" src="https://github.com/user-attachments/assets/2fb82115-b71e-4c74-b1c9-37294e6caf89" />

Leberkasbepi:
Same for me, but don't matter if I use another playlist settings.

phattmatt:
I can reproduce this bug as described.

Looks like the weirdness is happening when submitting a change to the plugin instance.

When the Moon Phase checkbox is checked (from unchecked) then saved, the value in `device.json` is `"moonPhase": "true"`, which works.

When making any other changes and saving, then the value in `device.json` is `"moonPhase": "on"`, which doesn't work.

phattmatt:
I think I've found the issue.

When a plugin instance settings page is loaded, the current settings for that instance are set from the values stored in `device.json`. The Moon Phase checkbox 'checked' value is correctly set, but the checkbox 'value' is not being set; I assume the default values for a checkbox 'value' are on/off, rather then true/false.

Compare:

https://github.com/fatihak/InkyPi/blob/0d21ceaf84e2ebe3d16ee44ffafb8e998b5a5bd8/src/plugins/weather/settings.html#L177

With a similar checkbox:

https://github.com/fatihak/InkyPi/blob/0d21ceaf84e2ebe3d16ee44ffafb8e998b5a5bd8/src/plugins/weather/settings.html#L173-L174

So I think we just need to add:

```javascript
document.getElementById('moonPhase').value = pluginSettings.moonPhase;
```

phattmatt:
The above change fixes the issue for me. I'll raise a small PR when I get a chance, unless someone beats me to it.

Leberkasbepi:
Works but 0%? 😅

<img width="1000" height="600" alt="Image" src="https://github.com/user-attachments/assets/62974511-a97d-4601-9c84-af63964f9804" />

[...]

PR #247, opened 2025-08-02
Add moonPhase value assignment to fix Moon Phase setting not working when saving plugin settings

phattmatt:
Small PR to fix the Moon Phase setting being set incorrectly when saving plugin instance settings.

Should fix the issue as reported here: https://github.com/fatihak/InkyPi/issues/245


fatihak (approved):

fatihak:
Looks good! Merged

Issue #239, opened 2025-07-28
Weather Plugin - Visability Issue - Not Respecting Saved Units of Measure


cello8619:
**Describe the bug**
Open Weather is still showing KM instead of Miles

**Expected behavior**
It should change between KM and MI depending on the saved units of measure.

fatihak:
This is an issue with the OpenWeatherMap API, looks like they always provide the visibility measurement in km regardless of the units parameter. However we could do the conversion in the weather plugin based on the units selected.

phattmatt:
Removing the bug tag, since this is not a bug with InkyPi, but rather a behaviour of the OpenWeatherMap API:

https://openweathermap.org/api/one-call-3#data:~:text=current.visibility%20Average%20visibility%2C%20metres.%20The%20maximum%20value%20of%20the%20visibility%20is%2010%20km

phattmatt:
This also seems the case for OpenMeteo API as well:

https://open-meteo.com/en/docs

Visibility is only returned in meters.

cello8619:
Pull Request - https://github.com/fatihak/InkyPi/pull/252 Submitted to add this enhancement.

jjacobowitz:
I've added pull request #482 to fix this.

fatihak:
I merged @jjacobowitz's PR, feel free to pull the latest changes and give it a try!

PR #252, opened 2025-08-06
Update Weather Plugin Visability Display to update based on the units of measure setting

cello8619:
Adding support to switch the visibility based on the unit of measure is selected.

This is update is for: Issue https://github.com/fatihak/InkyPi/issues/239


k2v1n58:
The visibility is already in ft with open-meteo, no need to update for it, only for OWM

```
OPEN_METEO_UNIT_PARAMS = {
    "standard": "temperature_unit=kelvin&wind_speed_unit=ms&precipitation_unit=mm",
    "metric":   "temperature_unit=celsius&wind_speed_unit=ms&precipitation_unit=mm",
    "imperial": "temperature_unit=fahrenheit&wind_speed_unit=mph&precipitation_unit=inch"
}
```
Here is an example: [API open-meteo imperial](https://api.open-meteo.com/v1/forecast?latitude=52.52&longitude=13.41&hourly=temperature_2m,visibility&precipitation_unit=inch)

jjacobowitz:
I'm interested in writing this fix. 

Currently, OpenWeather shows only km and open-meteo shows ft for imperial and km otherwise. I believe miles should be used for both OpenWeather and open-meteo when using imperial.

The change is simple for OpenWeather (simply convert from km to miles). For open-meteo, the code currently shows `>X` for visibility `X` greater than 10 km, but still shows the value for `X` since open-meteo does not appear to have a maximum visibility value. Is that still the desired behavior or should it be capped at 10 km (6.2 mi) to match the OpenWeather result?

One last suggestion: because OpenWeather is capped at 10 km, I propose using `u"\u2265"` (≥) for the symbol since the visibility may be greater than or equal to the limit.

I'll write up the code once we have a consensus on the changes to be made. This will be my first open source contribution, so I might need some help making sure I do it correctly.

PR #264, opened 2025-08-12
New moon icons

k2v1n58:
Update the moon icons and fix a bug on the 3rd quater when using open-météo and farmsense


fatihak (approved):

Issue #257, opened 2025-08-10
Plugins relying on 'chromium-headless-shell' sometimes fail with "Bus error"


phattmatt:
> The error I'm getting with chromium-headless-shell in the new build (Bus error) looks like memory error.  
> The screenshot is loaded _sometimes_, but usually crashes.
> log attached.
> 
> [inkypi_20250810-072528.log](https://github.com/user-attachments/files/21705177/inkypi_20250810-072528.log) 

 _Originally posted by @TonyTTTT in [#159](https://github.com/fatihak/InkyPi/issues/159#issuecomment-3172579353)_

I've also seen this error, specifically happens on the Weather Plugin, weirdly the Calendar Plug in (both rely on chromium-headless-shell) still seems to be fine.

I suspect it MAY be related to the recent Chromium release (2025-08-06).

Downgrading to a previous release of Chromium fixes this for me.

phattmatt:
To reproduce:

Check which version of Chromium packages you have installed (I think the latest release is '139.0.7258.66-1~deb12u1+rpt1' which is semi-broken for me) :

```shell
pi@inkypi-dev:~ $ dpkg-query -W chromium-headless-shell chromium-common
chromium-common 1:139.0.7258.66-1~deb12u1+rpt1
chromium-headless-shell 1:139.0.7258.66-1~deb12u1+rpt1
``` 

Try to update a Weather Plugin instance. You may get:

<img width="412" height="68" alt="Image" src="https://github.com/user-attachments/assets/0bccff54-9706-4fa2-8300-b9b0176fb727" />

Check the log files:

```
Aug 10 11:04:53 inkypi-dev inkypi[3401]: 11:04:53 - ERROR - utils.image_utils - Failed to take screenshot:
Aug 10 11:04:53 inkypi-dev inkypi[3401]: 11:04:53 - ERROR - utils.image_utils - [0810/110436.293253:WARNING:media/audio/linux/audio_manager>
Aug 10 11:04:53 inkypi-dev inkypi[3401]: Bus error
``` 

Workaround - Downgrade to previous release: 

```shell
pi@inkypi-dev:~ $ sudo apt install chromium-common=138.0.7204.183-1~deb12u1 chromium-headless-shell=138.0.7204.183-1~deb12u1
```

Try to update a Weather Plugin instance, it should now successfully refresh and display an image.

[...]

PR #265, opened 2025-08-12
Pin required Chromium Packages to "138.0.7204.183-1~deb12u1" as a workaround for issue #257

phattmatt:
An issue with version "139.0.7258.66-1~deb12u1+rpt1" of Chromium-Common and Chromium-Headless-Shell has been identified; a workaround of downgrading to the previous release has been confirmed.

The issue is being tracked here: https://github.com/fatihak/InkyPi/issues/257

This PR adds explicit versions in the `install/debian-requirements.txt` file which should ensure that only that version is installed during an install. If newer packages are already installed (i.e. during an update) they should be downgraded to the known working version.

### DIFF
--- a/install/debian-requirements.txt
+++ b/install/debian-requirements.txt
@@ -5,7 +5,8 @@ python3-pip
 avahi-daemon
 libopenblas-dev
 libopenjp2-7
-chromium-headless-shell
+chromium-common=138.0.7204.183-1~deb12u1
+chromium-headless-shell=138.0.7204.183-1~deb12u1
 libfreetype6-dev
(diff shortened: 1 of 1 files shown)


phattmatt:
I've tested this change in a few different scenarios and it seems to behave well:

1. Fresh install, no packages install - pinned version is installed
2. Update on already pinned version - pinned version is kept the same
3. Update on newer version - version is downgraded to pinned version
4. Install on newer version - version is downgraded to pinned version

Marking as ready for review.

fatihak (approved):

PR #270, opened 2025-08-15
Image folder plugin

zapper1942:
Description:
New plugin to pull images from a local folder and display them in a random order. 
The main intent behind this was to just point to a local folder filled with lots of pictures I can update instead of manually uploading a bunch of individual image files.
It's a bit more technical but I have the local Pictures folder mounted to a shared folder on my NAS. In addition to a centralized place the put pictures on, this should also allow multiple frames to pull the same images.

The size restrictions are because excessively large files would overload the swap file and crash the server.

Solution:
image
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ InkyPi is an open-source, customizable E-Ink display powered by a Raspberry Pi. 
 
 **Plugins**:
 
-- Image Upload/Local Folder: Upload and display any image from your browser or mount a local Folder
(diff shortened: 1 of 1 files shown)


zapper1942:
Now I see that you already implemented this, I guess I should've synced sooner. I'll level this up in case there are aspects of my implementation that you want to merge in.

fatihak:
Closing Draft PR, feel free to reopen when its ready to merge

PR #271, opened 2025-08-15
Fix system dependencies install on pristine OS

neversun:
On a pristine OS install (in my case: Raspberry PI OS Lite)  there might never have been run `apt-get update`.
But without a previous run, the following `apt-get install $debian_deps_here$` will _most likely_ fail.

This fixes it by running apt-get update before apt-get install.


fatihak (approved):
Looks good, thanks for the PR

PR #275, opened 2025-08-17
Add amount of rain

k2v1n58:
This PR offers the possibility to write the amount of rain hour by hour on the graph.
It works with OWM, OM and for metric and imperial units.

To be tested

image


fatihak:
Looks good, was able to test with both Open Weather Map and Open Meteo and saw the rain amounts populated correctly. Thanks for the PR!

fatihak (approved):

Issue #278, opened 2025-08-19
InkyPi says Waveshare display does not support required methods of the corresponding model


TomLisankie:
**Describe the bug**
I am using [this](https://www.waveshare.com/wiki/13.3inch_e-Paper_HAT+_(E)_Manual) Waveshare display. It is one of the models listed in the alternative directory in Waveshare's ePaper repository and is the same exact one used as an example in [InkyPi's troubleshooting guide](https://github.com/fatihak/InkyPi/blob/main/docs/troubleshooting.md#waveshare-e-paper-epd-devices). I followed the instructions in that section of the troubleshooting guide exactly and the InkyPi service still always dies after just a few seconds.

_**I know for a fact that the display itself is not the problem since I ran Waveshare's test script for the corresponding display and all of the graphics the test script created displayed just fine.**_

**Expected behavior**
I expected the InkyPi splash screen to display after following the Waveshare instructions in the troubleshooting guide and following the instructions on what to do after a clean install.

**To Reproduce**
1. Properly connect this [this](https://www.waveshare.com/wiki/13.3inch_e-Paper_HAT+_(E)_Manual) Waveshare display to a Raspberry Pi Zero 2 W.
2. Clone InkyPi onto the Pi and follow the instructions to install along with the troubleshooting steps relevant to that particular Waveshare display.
3. Follow post-instruction steps.
4. Notice how what is supposed to happen is not happening.

**Logs**
```
Aug 18 23:37:36 portrait systemd[1]: Started inkypi.service - InkyPi App.
Aug 18 23:37:50 portrait inkypi[663]: /usr/local/inkypi/venv_inkypi/lib/python3.11/site-packages/inky/__init__.py:18: UserWar>
Aug 18 23:37:50 portrait inkypi[663]:   from pkg_resources import declare_namespace
Aug 18 23:37:52 portrait inkypi[663]: 23:37:52 - INFO - __main__ - Starting web server
Aug 18 23:37:52 portrait inkypi[663]: 23:37:52 - INFO - display.waveshare_display - Initializing Waveshare display
Aug 18 23:37:52 portrait inkypi[663]: 23:37:52 - INFO - display.waveshare_display - Loading EPD display for epd13in3E display
Aug 18 23:37:52 portrait inkypi[663]: EPD init...
Aug 18 23:37:52 portrait inkypi[663]: Traceback (most recent call last):
[...]

Issue #188, opened 2025-07-02
What is the device name for the Waveshare Spectra 6 (E6) Full Color 13.3 inch?


evanthomas:
I can't find anything in 
```
https://github.com/waveshareteam/e-Paper/tree/master/RaspberryPi_JetsonNano/python/lib/waveshare_epd
```
corresponds to it.

fatihak:
Looks like the 13.3 inch E6 display driver is under:
https://github.com/waveshareteam/e-Paper/tree/master/E-paper_Separate_Program/13.3inch_e-Paper_E/RaspberryPi/python/lib/

Checkout the instructions [here](https://github.com/fatihak/InkyPi/blob/main/docs/troubleshooting.md#error-failed-to-download-waveshare-driver), you'll need to manually download the files.

evanthomas:
The installation appeared to run correctly:
```
pi@art-display:~/InkyPi $ sudo bash install/install.sh -W epd13in3E
Optional parameter WS is set for Waveshare support.  Screen type is: epd13in3E
Checking if inkypi.service is running
	inkypi.service not running [✔]
Fetching Waveshare driver for: epd13in3E
	Waveshare driver 'epd13in3E.py' already exists at /home/pi/InkyPi/install/../src/display/waveshare_epd/epd13in3E.py [✔]
	Waveshare epdconfig file already exists at /home/pi/InkyPi/install/../src/display/waveshare_epd/epdconfig.py [✔]
Enabling interfaces required for inkypi
	SPI Interface has been enabled. [✔]
	I2C Interface has been enabled. [✔]
Enabling both CS lines for SPI interface in config.txt
Extracting templates from packages: 100%
Installing system dependencies.  [✔] 
Installing inkypi to /usr/local/inkypi
	Creating symlink from /home/pi/InkyPi/install/../src to /usr/local/inkypi/src [✔] 
Creating python virtual environment. 
	Installing python dependencies.  [-] WARNING: Retrying (Retry(total=4, connect=None, read=None, redirect=None, status=None)) after connection broken by 'ProtocolError('Connection aborted.', RemoteDisconnected('Remote end closed connection without response'))': /simple/pillow/
	Installing python dependencies.  [|] WARNING: Retrying (Retry(total=4, connect=None, read=None, redirect=None, status=None)) after connection broken by 'ProtocolError('Connection aborted.', RemoteDisconnected('Remote end closed connection without response'))': /simple/openai/
	Installing python dependencies.  [✔] 
[...]

PR #279, opened 2025-08-19
Update troubleshooting.md

giovanniborella:
Add missing file to download when using the waveshare epd13in3E

ref: #278 #188


TomLisankie:
Can confirm that this fixes the issue. However, I think the instructions should mention that the DEV_config you opt to download depends on the Raspberry Pi device you have, not just the Waveshare display. For example, if someone was following those instructions and they had a Raspberry Pi 5, they would be downloading the wrong object file and it might not work.

giovanniborella:
@TomLisankie that is great info - do you know any of the other files? I only have a zero 2w on hand

fatihak (approved):

PR #280, opened 2025-08-19
Update weather.py for visibility correct units

cello8619:
feat(weather): normalize visibility to km/mi for OpenWeather + Open-Meteo

- Open-Meteo: read hourly_units.visibility (m/ft) and convert to target display unit based on user preference (imperial→mi, metric/standard→km).
- OpenWeather: convert current.visibility (meters) to km/mi for display.
- Apply caps to match common reporting: >10.0 km / >6.2 mi (10 km ≈ 6.2137 mi).
- Friendly rounding: 1 decimal (whole numbers for very large values).
- Ensure unit label always matches user preference; remove prior 'ft' display path.
- Handle missing fields gracefully ("N/A"); improved logging for parse/index issues.
- Match current local hour; fallback to first hourly value when exact match not found.
- Add helpers: convert_distance() and display_visibility_from_raw() for reuse.

No external API contract changes; only display/formatting made consistent.

### DIFF
--- a/src/plugins/weather/weather.py
+++ b/src/plugins/weather/weather.py
@@ -1,642 +1,273 @@
+# src/plugins/weather/weather.py
+
 from plugins.base_plugin.base_plugin import BasePlugin
-from PIL import Image
-import os
-import requests
 import logging
-from datetime import datetime, timezone
-import pytz
-from io import BytesIO
 import math
+import os
+from datetime import datetime
+from io import BytesIO
+from typing import Optional, Tuple, Dict, Any, List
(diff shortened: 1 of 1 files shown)


k2v1n58:
I have the feeling that your code removes many functionalities, like the moon phase, AQI and icons management.
The API from open-meteo returns the right unit depending if metric or imperial is selected on the settings.

fatihak:
Rejecting this PR as it reworks the entire plugin for a simple change and seems AI generated.

PR #286, opened 2025-08-24
fix: plugin icon serving

florianbachmann:
# Fix Plugin Icon Serving

## Problem

Plugin icons return `404` errors when InkyPi is started from different directories. The module-level `PLUGINS_DIR` variable is resolved at import time, which can point to the wrong location depending on the working directory at startup.

## Solution

Resolve plugin paths dynamically within request handlers instead of at module import time. This ensures icons load correctly regardless of where InkyPi is started from.

## What Changed

**File:** `src/blueprints/plugin.py`

- Removed module-level `PLUGINS_DIR` variable
- Added dynamic path resolution in the image route handler
- Added security validation to prevent directory traversal attacks
- Added proper error handling with logging for debugging

## Impact

+ Fixes plugin icons returning `404` errors  
+ Works regardless of InkyPi's working directory at startup  
+ Improves security with path validation  
+ Better error logging for troubleshooting

### DIFF
--- a/src/blueprints/plugin.py
+++ b/src/blueprints/plugin.py
@@ -9,7 +9,7 @@ import logging
 logger = logging.getLogger(__name__)
 plugin_bp = Blueprint("plugin", __name__)
 
-PLUGINS_DIR = resolve_path("plugins")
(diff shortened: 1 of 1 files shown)


fatihak (approved):

PR #298, opened 2025-09-03
feat: Comic enhancements

olma2077:
Multiple comic plugin enhancements:

- Features:
  - Added "webcomic name" feed.
  - Upscale image for small comic panels (looking at you, xkcd!) or big screens.
  - Optional comic title and caption when present.
  - Possibility to select title/caption font size.
- Refactor:
  - Feed parser in a separate module.
  - Feed parsing is linked with the comic list, no way to forget adding parsing logic when adding a new feed.
  - Image composition in a separate method as logic became more complex.

### DIFF
--- a/src/plugins/comic/comic.py
+++ b/src/plugins/comic/comic.py
@@ -43,20 +43,24 @@ class Comic(BasePlugin):
                 if comic_panel["title"]:
                     lines, wrapped_text = self._wrap_text(comic_panel["title"], font, width)
                     draw.multiline_text((width // 2, 0), wrapped_text, font=font, fill="black", anchor="ma")
-                    top_padding = font.getbbox(wrapped_text)[3] * lines
+                    top_padding = font.getbbox(wrapped_text)[3] * lines + 1
 
                 if comic_panel["caption"]:
                     lines, wrapped_text = self._wrap_text(comic_panel["caption"], font, width)
                     draw.multiline_text((width // 2, height), wrapped_text, font=font, fill="black", anchor="md")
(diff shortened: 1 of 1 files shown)


fatihak:
Really like the refactoring of the feed parsing logic. Thanks for the PR!

A few issues:
- For Dinosaur Comics I dont think the title is parsed correctly, its set to "this comic brought to you by Dinosaur Comics brand dinosaur comics"
- Questionable Content has a title in the feed but its not being pulled
- "DejaVuSans.ttf" isnt committed to the repo so I don't think its being used. You can either use one of the existing fonts in src/static/fonts/Jost.ttf or commit the font to the repo
- Sometimes the comic cuts off the caption below, here is an example with xkcd
current_image (1)

olma2077:
@fatihak, thank you for the feedback! Glad you liked the refactor :)

1. It is [the real title](https://www.qwantz.com/archive.php) for yesterday's panel, the guy has a peculiar sense of humor.
2. You're right, somehow missed it. Fixed.
3. DejaVuSans is in the base RaspberryPiOS, PIL uses fonts available through `fc-list`. Jost is not installed in the system with your project, so either `install.sh` should be updated to install and then use it, or a default available font in an OS could be used. I'd keep the system font.
4. ~Which device are you using? I have only 7.3 Pimoroni, it renders ok. From your screenshot, it seems like the image shifts down a bit.~

olma2077:
Follow up on the 4th item: apparently, it depends on the font size, looking into it.

olma2077:
Alignment redone.
Regarding font usage, I'd stick with the system one.

PR #304, opened 2025-09-10
Image folder update

RoliMG:
add option to prevent image repetition
add option for panorama and portrait folder for image selection
add shutdown after image display option


anton-ovcharenko:
Hi @RoliMG, why do you need shutdown functionality?

RoliMG:
my raspberry changes the image once per day and turns off after the image was updated. didn't find a feature in inkypi that would allow an automated shutdown.

fatihak:
Thanks for the PR, but I don't think shutting down after displaying an image is a necessary feature. Also, having separate horizontal/vertical directories doesn’t seem necessary—the display orientation is typically fixed, and the current logic already handles images correctly based on the chosen setting.

Issue #211, opened 2025-07-12
Immich API Integration


exxrxxe:
Would be cool to see a Immich (http://immich.app) API integration so photos could be managed on an Immich server and display on InkyPi.

fatihak:
At the moment, there aren’t any plans to support the Immich API directly, since there are so many different photo management platforms and it wouldn’t be sustainable to maintain dedicated plugins for each. That said, you might be able to build a custom solution using the Image URL plugin by pointing it to publicly accessible image links from your Immich server.

fatihak:
Just merged a PR by @RoliMG that adds an Image Album plugin with Immich API Integration. Feel free to give it a try!

Sawtaytoes:
It says "Requires API Key", but I don't see a place to add one:
<img width="997" height="379" alt="Image" src="https://github.com/user-attachments/assets/4dfa3c89-e0bc-48a9-92e0-8dd19775ba1c" />

ivignesh:
I followed the instructions in this doc:
https://github.com/fatihak/InkyPi/blob/main/docs/api_keys.md

I added my Immich API key to the .env file and restarted the Pi. However, when I try to add the album in the web UI with the exact album name, I still get this error:
Error! An error occurred: Immich API Key not configured.

Any guidance on why it’s not recognizing the key?

RoliMG:
> I followed the instructions in this doc: https://github.com/fatihak/InkyPi/blob/main/docs/api_keys.md
> 
> I added my Immich API key to the .env file and restarted the Pi. However, when I try to add the album in the web UI with the exact album name, I still get this error: Error! An error occurred: Immich API Key not configured.
> 
> Any guidance on why it’s not recognizing the key?

Guidance: [https://github.com/fatihak/InkyPi/tree/main/src/plugins/image_album](https://github.com/fatihak/InkyPi/tree/main/src/plugins/image_album)

ivignesh:
After further investigation, I identified the issue — it was caused by an incorrect BASE URL provided on my end. Once corrected to use only the server root (without the album path), the Immich album loaded correctly. Everything is working as expected now.

Really appreciate the work on InkyPi. Closing this issue from my side 👍

PR #321, opened 2025-10-07
Immich album plugin

RoliMG:
loads images from an album on an Immich server
support for heif image format added
still missing: image settings via the UI (padding, fit to frame, etc)

### DIFF
--- a/src/plugins/base_plugin/base_plugin.py
+++ b/src/plugins/base_plugin/base_plugin.py
@@ -2,14 +2,12 @@ import logging
 import os
 from utils.app_utils import resolve_path, get_fonts
 from utils.image_utils import take_screenshot_html
-from pillow_heif import register_heif_opener
 from jinja2 import Environment, FileSystemLoader, select_autoescape
 from pathlib import Path
 import asyncio
 import base64
 
 logger = logging.getLogger(__name__)
-register_heif_opener()
 
 STATIC_DIR = resolve_path("static")
 PLUGINS_DIR = resolve_path("plugins")


F2011:
This feature was already rejected in #211. Although, I think it is a good new plugin. Immich is by far the biggest Open-Source Photo Management solution, and it makes totally sense to provide a plugin for it.

Maybe, it is possible to make this a bit agnostic and make it easily extendible for other photo platforms, somehow like the weather plugin where you can switch between OpenMeteo and OpenWeatherMap.

deftdawg:
Immich being my primary use case for a digital frame, I hope this can be merged.

src/plugins/image_album/image_album.py
```diff
@@ -0,0 +1,139 @@
...
+        r = requests.get(f"{self.base_url}/api/albums", headers=self.headers)
+        r.raise_for_status()
+        albums = r.json()
+        album = [a for a in albums if a["albumName"] == album][0]
+        return album["id"]
+
+    def get_asset_ids(self, album_id: str) -> list[str]:
+        body = {
+            "albumIds": [album_id],
+            "size": 1000,
```
fatihak: This will only return the first 1000 assets for large albums, we should paginate through responses using the nextPage property to get all assets

src/utils/image_utils.py
```diff
@@ -147,3 +147,12 @@ def take_screenshot(target, dimensions, timeout_ms=None):
         logger.error(f"Failed to take screenshot: {str(e)}")
 
     return image
+
+def pad_image_blur(img: Image, dimensions: tuple[int, int]) -> Image:
```
fatihak: This function is already in image_utils, please update the PR with the latest changes from main

src/plugins/image_album/settings.html
```diff
@@ -0,0 +1,55 @@
...
+</div>
+
+<div class="form-group">
+    <label for="randomize" class="form-label">Randomize Order:</label>
+    <div class="toggle-container">
+        <input type="checkbox" id="randomize" name="randomize" class="toggle-checkbox" value="false"
+            onclick="this.value = this.checked ? 'true' : 'false'">
+        <label for="randomize" class="toggle-label"></label>
+    </div>
+    <label for="randomRepetition" class="form-label">Allow repeating randomly selected images</label>
```
fatihak: I dont think this is needed, random usually means it can be repeated and this makes it confusing

RoliMG:
Thanks for the comments, will have a look!

iAndrea:
Awesome! I'll update my branch https://github.com/fatihak/InkyPi/pull/427 to enable optimizations for this new `image_album` plugin 😃

Sawtaytoes:
@RoliMG Howdy! This is a fantastic addition to InkyPi! I was wondering if it's possible to create a dynamic album?

For instance, I'd like to search "Kids", set a start date, and search only "images":
image

The URL looks like this:
```
/search?query=%7B"query"%3A"kids"%2C"takenAfter"%3A"2018-01-01T00%3A00%3A00.000Z"%2C"type"%3A"IMAGE"%7D
```

If I could create a dynamic album based off a filtered search, then it can dynamically pull the photos. Right now, I have to manually select them all, add them to an album, and when I get new photos, I need to do it again.

RoliMG:
Hi @Sawtaytoes! You mean that you would like to basically display images from a search?

PR #330, opened 2025-10-20
Fix: Rename method image_blur

Rashgaron:
The method had the wrong name.
So when selecting Fit the image to the frame on image folder options, it wasnt working


fatihak:
Whoops, good catch :)

PR #336, opened 2025-10-21
Add "Word of the Day" plugin with API integration and OpenAI processing

Rashgaron:
This PR introduces a "Word of the Day" plugin capable of fetching a random word from an external API. 

Key features:
- Retrieves a random word via API.
- Uses OpenAI to generate additional details and translations based on the selected language.
- Displays the word, part of speech, description, and example usage.
- Fully supports multiple languages and dynamic content generation.

This plugin provides a flexible and interactive way to enhance language learning and vocabulary exposure.

### DIFF
--- a/src/plugins/word_of_the_day/word_of_the_day.py
+++ b/src/plugins/word_of_the_day/word_of_the_day.py
@@ -87,18 +87,19 @@ class WordOfTheDay(BasePlugin):
         system_content = "You are an assistant that returns only valid JSON objects."
 
         user_content = f"""
-        Given the word '{word}', provide its equivalent in {text_lang}.
+        Given the word '{word}', provide its most natural equivalent in {text_lang}.
+        If the language commonly uses a native term instead of a loanword, prefer the native word even if the meaning is not a perfect one-to-one match.
+        If no natural native equivalent exists, use the closest borrowed or mixed form.
 
         Return a valid JSON object with the following fields:
-        - word: the translated word in {text_lang}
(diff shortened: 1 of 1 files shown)


fatihak:
Thanks for the PR! This is a pretty nice plugin but I’m not a big fan of using OpenAI for the processing since a lot of the same functionality can already be achieved using the existing AI Text plugin. I’d be more open to merging this if it could rely on free APIs or libraries instead.

For example, I was able to produce something very similar using the AI Text plugin with the following prompt, which can be further tuned for formatting and design:

```
Give me a random word and provide its most natural equivalent in Spanish.
If the language commonly uses a native term instead of a loanword, prefer the native word even if the meaning is not a perfect one-to-one match.
If no natural native equivalent exists, use the closest borrowed or mixed form.
Return the following as a formatted html that will be substituted inside of an existing html div. Each piece of information should be in a <p> tag. Only use inline css, do not add additional css classes or external styles. Do NOT return Markdown code blocks. Just return plain HTML/text.
- word: the most natural or native equivalent in Spanish
- type: part of speech (noun, verb, adjective, etc.)
- meaning: a concise definition in English
- example: one clear example sentence showing correct usage (in Spanish)
- lecture: include ONLY if the word is not written in the Roman alphabet. Provide the correct pronunciation in a readable phonetic form (e.g., hiragana/katakana for Japanese, pinyin for Chinese, transliteration for other scripts).
Do NOT return the original script.
```

Resulting image:
word_of_the_day_test

PR #339, opened 2025-10-22
Check for OS version and only install zramswap on Bookworm

phattmatt:
Raspberry Pi OS 13 (Trixie) has been released.

Trixie includes a zram based swap out of the box:

https://forums.raspberrypi.com/viewtopic.php?t=390708

The zram-tools package the InkyPi 'install.sh' installs conflicts with the pre-installed 'rpi-swap' package.

This PR modifies the 'install.sh' file by adding a OS version check and only installing/configuring zram-tools/zramswap on Raspberry Pi OS 12 (Bookworm).

### DIFF
--- a/install/debian-requirements.txt
+++ b/install/debian-requirements.txt
@@ -8,5 +8,3 @@ libopenjp2-7
 chromium-headless-shell
 libfreetype6-dev
 fonts-noto-color-emoji
-zram-tools
-earlyoom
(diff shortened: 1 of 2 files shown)


fatihak (approved):

fatihak:
Looks good, thanks for the fix

TonyTTTT:
Installation correctly checks:
OS version is not Bookworm - skipping zramswap setup.
But Update does not:
Starting zramswap service.
Failed to enable unit: Unit zramswap.service does not exist

PR #373, opened 2025-11-02
feat: bundle JS and CSS instead of using CDN

### LINKED ISSUE #358: Idea: Bundle JavaScript and CSS instead of using CDN
Hi, 

I would love an option that made it possible to bundle the CSS and JavaScript instead of using a CDN. This would make it GDPR Conforming too, in that regard.  I prefer not to make requests out to the world more than needed for a device like this. 

Update: 

Not alone will it GPDR conforming, but it will also be faster, as the assets will not be loaded from external resources.

tomasnorre:
Resolves #358 

This is a simple first draft, that I would like some feedback on. It's quite simplified. 

I'm considering following
* renaming e.g. `src/static/scripts/jquery-3.6.0.min.js` -> `src/static/scripts/jquery.min.js` to only have to update the `scripts/update_vendors.sh` and not the `templates/plugins.html` too.
* using a package manager like `npm` or even leveraging a tool like `ViteJS`

What is your opinion?


fatihak:
I think this is a good approach, I prefer the first option since it’s more lightweight rather than installing additional packages

A couple of suggestions:
- We should call the update_vendors.sh from install.sh and update.sh so they're automatically fetched.
- Some plugins like Weather and Calendar also rely on CDNs, those could be added here as well for consistency.

tomasnorre:
Is must have overlooked them as searching for JS and CSS. I'll update the PR

tomasnorre:
I think it's ready for review now.

Will suggest a squash and merge when it's we are so far.

fatihak:
Do we need the actual files committed to the repo since we're already pulling them during install & update?

tomasnorre:
> Do we need the actual files committed to the repo since we're already pulling them during install & update?

As neither `install/install.sh` nor `install/upgrade.sh` is called on local development, https://github.com/fatihak/InkyPi/blob/main/docs/development.md, we would either need to include them, the files, or instruct in the documentation how to fetch them for a development environment. 

I think the best way is to exclude them from the repository, as it will require less maintenance. If we exclude them, they should also be added to `.gitignore` to avoid them being committed by mistake later.

fatihak (approved):

PR #375, opened 2025-11-02
feat: add GitHub Sponsor plugin

tomasnorre:
I have created a GitHub Sponsors plugin, it currently only shows the "earning this month" but could be extended to show top 3 sponsors or similar.

An improvement that just came up, was the possibility to show "30% of goal reached" with a progress bar. But that should be an option. 

Until we have a "3rd plugin" implementation setup, I'll keep suggesting my plugins as PRs :)


fatihak:
I think both of your Github related plugins can be combined in to the existing Github plugin, with an option to choose between Commit Graph, Sponsors, or Stars. I would be fine with merging both if that were the case.

tomasnorre:
Sounds like a good idea to combine them. I'll take care of that.

tomasnorre:
Closed, as this will be combined with existing GitHub Plugin, will create a new PR.

PR #377, opened 2025-11-03
feat: add GitHub Sponsors and Repository Stars view

tomasnorre:
This PR combines the two previous PRs that added need plugins for GitHub Stars and GitHub Sponsors. 

Currently, it's only the GitHub Sponsors that are included, but not possible to select.

I wanted to hear your opinion on the way I have split this up into separate files, for "single responsibility"-purpose and easier maintenance. 

So let me hear what you think of this first step, then I'll move forward. I'll improve the sponsoring with progress too, but firstly I want to hear the overall opinion about the way I see this change.


fatihak:
Yeah this is the right approach, keeps everything clean and organized

PR #386, opened 2025-11-06
Adds a plugin that grabs a random image from a Google Photos album share link

mohlendo:
The Plugin does not need any authentication since it only uses a public google photos album. There is a little bit of html/js parsing which might break some time. The json5 lib is needed because it parses Javascript objects and not valid json.


fatihak:
Parsing HTML and JavaScript is fragile and hard to maintain, as small changes can easily break the code. I’d be more open to merging a plugin that uses the API instead.

Issue #370, opened 2025-11-01
[Feature Request] Allow Image Folder Plugin to include SubDirectories


TonyTTTT:
Might be worth allowing the Image Folder Plugin to include subdirectories.

For example, I have the following structure:
`-- Pictures
    |-- Astronomy Picture of the Day
    `-- Favorites

This can be accomplished by adding to image_folder.py:

from pathlib import Path

def list_files_recursively(folder_path, extensions=None):
    """
    Recursively lists all files in folder_path and subfolders.
    Optionally filters by file extensions.
    """
    folder = Path(folder_path)
    if extensions:
        extensions = tuple(ext.lower() for ext in extensions)
        return [str(p) for p in folder.rglob("*") if p.suffix.lower() in extensions]
    else:
        return [str(p) for p in folder.rglob("*") if p.is_file()]

And changing:
image_files = list_files_in_folder(folder_path)
to
image_files = list_files_recursively(folder_path, extensions=[".jpg", ".jpeg", ".png", ".bmp", ".gif"])

roberthunt:
This should now be resolved if you pull in the latest changes as it was fixed in https://github.com/fatihak/InkyPi/pull/388

## Update
To update your InkyPi with the latest code changes, follow these steps:
1. Navigate to the project directory:
    ```bash
    cd InkyPi
    ```
2. Fetch the latest changes from the repository:
    ```bash
    git pull
    ```
3. Run the update script with sudo:
    ```bash
    sudo bash install/update.sh
    ```
This process ensures that any new updates, including code changes and additional dependencies, are properly applied without requiring a full reinstallation.

fatihak:
Closing this as this is now available from @roberthunt's merged PR.

PR #388, opened 2025-11-07
Add support for including subfolders when listing image files in "Image Folder" plugin

roberthunt:
Adds a new toggle to enable scanning of subfolders for images (defaults to off to retain existing behaviour). Addresses feature request from https://github.com/fatihak/InkyPi/issues/370

Also updated log statement to indicate if subfolders are being included or excluded.

```
22:03:14 - INFO - plugins.image_folder.image_folder - Grabbing a random image from: /Users/rob/Downloads/root (excluding subfolders)
...
22:03:24 - INFO - plugins.image_folder.image_folder - Grabbing a random image from: /Users/rob/Downloads/root (including subfolders)
```

Screenshot 2025-11-07 at 22 08 19

### DIFF
--- a/src/plugins/image_folder/image_folder.py
+++ b/src/plugins/image_folder/image_folder.py
@@ -8,7 +8,7 @@ from utils.image_utils import pad_image_blur
 
 logger = logging.getLogger(__name__)
 
-def list_files_in_folder(folder_path, include_subfolders):
+def list_files_in_folder(folder_path):
     """Return a list of image file paths in the given folder, excluding hidden files."""
     image_extensions = ('.jpg', '.jpeg', '.png', '.gif', '.bmp', '.tiff', '.webp')
     image_files = []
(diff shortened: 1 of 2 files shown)


fatihak:
I don't think we need an option in the UI, we can just always include subfolders when listing images

roberthunt:
> I don't think we need an option in the UI, we can just always include subfolders when listing images

Yeah, I did approach it this way initially, but thought that some existing users might get a surprise if the behaviour changes.

I do think that scanning the subfolders would be the expected behaviour of this plugin. I'm happy to modify the PR to remove the UI changes if you would like?

fatihak:
Yeah i think its fine to always scan subfolders

roberthunt:
@fatihak Changes pushed, removed all of the UI changes and parameters for subfolders

fatihak (approved):

PR #399, opened 2025-11-10
Make the newspaper and wptod plugins orientation aware

deftdawg:
When I set my device to vertical, the newspaper plugin was still rendering landscape.   Flipping Height and Width resolved it for newspaper but broke it for all other plugins.   The solution was to make newspaper orientation aware like the clock plugin.

**Update:** same done for wpotd when shrink is enabled, since it was sizing the image using vertical dimension for horizontal when device was in vertical orientation.


deftdawg:
@fatihak orientation bug fixes ready to merge

fatihak (approved):

PR #400, opened 2025-11-10
Add ability to pick chromium from device config or list of known good candidates

efrecon:
On the latest Fedora, I had problems finding the proper chromium binary to be able to render in headless mode. So, since there was already some documentation to manually change the code for Windows and MacOS, I made this a key that can be set in the `device.json` file (well, probably most useful for `device_dev.json`!).

### DIFF
--- a/docs/building_plugins.md
+++ b/docs/building_plugins.md
@@ -130,10 +130,19 @@ For more complex plugins or dashboards that display dynamic content, you can gen
 ### Using `render_image`
 You can generate an image by calling the `BasePlugin`'s `render_image` function, which accepts the following arguments:
 - `dimensions` (tuple)                  The width and height of the generated image.
-- `chromium` (str)                      The path to the chromium headless shell, defaults to `chromium-headless-shell`.
 - `html_file` (str)                     Name of the HTML file to render, located in the `render/` directory.
 - `css_file` (str, optional)            Name of the CSS file in the `render/` directory.
 - `template_params`(dict, optional)     A dictionary of values to be passed into the Jinja template.
+- `chromium` (str, optional)            The path to the chromium headless shell, see below for defaults.
+
+Plugins should pick the path to `chromium` from the key called `chromium` in the device configuration.
+When `chromium` is not found in the device configuration or not passed to `render_image`, it will pick the first of the following defaults available on the path:
+
+- `chromium-headless-shell`
+- `chromium`
+- `chromium-browser`
+- `google-chrome`
+- `chrome`
 
 ### Defining HTML and CSS files
 - Place your HTML and CSS files in the `render/` subdirectory, as `render_image` looks for them there.
--- a/src/plugins/ai_text/ai_text.py
+++ b/src/plugins/ai_text/ai_text.py
@@ -49,7 +49,7 @@ class AIText(BasePlugin):
         if device_config.get_config("orientation") == "vertical":
             dimensions = dimensions[::-1]
 
-        chromium = device_config.get_config("chromium", "chromium-headless-shell")
+        chromium = device_config.get_config("chromium")
 
         image_template_params = {
             "title": title,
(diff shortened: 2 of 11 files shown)


deftdawg:
I'm on NixOS, I've felt this challenge myself.   Below is an excerpt from an upcoming PR to enable development via  [devbox](https://www.jetify.com/devbox) on all Linux/WSL variants, maybe give it a whirl.  

It auto-detects for `chrome-headless-shell` (not available in nixpkgs), `chromium`, and `chrome` in the PATH, and runs with the first one it finds.  

```diff
diff --git a/src/utils/image_utils.py b/src/utils/image_utils.py
index b8cf0c0..60492e2 100644
--- a/src/utils/image_utils.py
+++ b/src/utils/image_utils.py
@@ -6,6 +6,7 @@ import logging
 import hashlib
 import tempfile
 import subprocess
+import shutil
 
 logger = logging.getLogger(__name__)
 
@@ -100,15 +101,32 @@ def take_screenshot_html(html_str, dimensions, timeout_ms=None):
 
     return image
 
+def _find_chromium_binary():
+    """Find the first available Chromium-based binary in system PATH."""
+    candidates = ["chromium-headless-shell", "chromium", "chrome"]
+    for candidate in candidates:
+        path = shutil.which(candidate)
+        if path:
+            logger.debug(f"Found browser binary: {candidate} at {path}")
+            return candidate
+    return None
+
+
 def take_screenshot(target, dimensions, timeout_ms=None):
     image = None
     try:
+        # Find available browser binary
+        browser = _find_chromium_binary()
+        if not browser:
+            logger.error("No Chromium-based browser found. Install chromium, chromium-headless-shell, or chrome.")
+            return None
+
         # Create a temporary output file for the screenshot
         with tempfile.NamedTemporaryFile(suffix=".png", delete=False) as img_file:
             img_file_path = img_file.name
 
         command = [
-            "chromium-headless-shell",
+            browser,
             target,
             "--headless",
             f"--screenshot={img_file_path}",
@@ -132,8 +150,7 @@ def take_screenshot(target, dimensions, timeout_ms=None):
 
         # Check if the process failed or the output file is missing
         if result.returncode != 0 or not os.path.exists(img_file_path):
-            logger.error("Failed to take screenshot:")
-            logger.error(result.stderr.decode('utf-8'))
+            logger.error(f"Failed to take screenshot (return code: {result.returncode})")
             return None
 
         # Load the image using PIL
```

deftdawg:
My [devbox](https://www.jetify.com/devbox) PR is up: https://github.com/fatihak/InkyPi/pull/404

You might check it out, since it should auto-configure on systems like Fedora (in addition to Arch, Debian, NixOS, MacOS, etc)

efrecon:
@deftdawg I think our PRs serve two slightly different goals: some devs might not want to install the entire devbox just for the sake of getting into the code. They will however probably bump into problems when pointing at a proper binary for `chromium`. Looking for good candidates on the path was a good idea, so I picked that and modified my code. I still think that it's a good idea to be able to point at a specific location through the device configuration. So my proposed (and slightly revised) logic for `chromium` is as follows, as of e97c6297dce2cbec29dcb4357dc54a93592e1a66:

- Plugins should try picking the location of `chromium` from the device configuration.
- `chromium` is now an optional parameter to `render_image`, so that it is ok for plugins not to pick (or to rely on the defaults).
- `render_image` will pick from a list of candidates whenever necessary.
- I have added yet another binary name (the one I needed on Fedora). This gives us an ok list to cover a range of distros.

You should still be able to build upon this inside your PR.

deftdawg:
> I still think that it's a good idea to be able to point at a specific location through the device configuration.

I don't really follow why though, like is the case here for folks that don't want chromium in their PATH?  They could override the PATH to include it before starting Inky... 

The other thing I don't understand is if you're going to have chromium as a device setting (rather than plugin specific), why do you need to load it into each plugin and then pass it to the renderer?   Why not just load it into the renderer and leave the plugin code alone?   Is there a case where you envision non-PATH chromium being used to render some plugins and PATH chromium being used to render others?  Such a case could make for confusing debug scenarios.

efrecon:
(sorry for long silence) was sick.

You are right, this is too complicated. Let me kill this and propose something simpler.

efrecon:
Created #430 as a replacement for this.

Issue #215, opened 2025-07-14
Feature Request: Split to client server architecture


sebaroesch:
First of all: thank you for creating and expanding this awesome project and the tutorials that go along with it! ♥️

I will definitely build this, just finishing my parts list. Back to topic:

InkyPi is great, but the overall setup forces the display to be powered by wire. It is to power hungry for a useful battery-powered setup.
Lots of people seem to be asking for a way to run this on battery. And with https://github.com/sjnims/rpi_weather_display there is even a similar project that achieves displaying weather on an e-ink screen while being powered from battery for weeks/months.

Suggestion:
Check if a client/server architecture is feasible for InkyPi and what steps are necessary.

What do you think?

knicefire:
Had similar thinking myself. 

Also would be nice for future integrations like Home Assistant, where HA does the heavy tasks and just pushes the images to the frame. 

Also a good thing for multiple frame orchestration scenario 

But, frankly, for HA integration nothing is stopping us to use existing solution as there is a convenient  `update_now`  endpoint that can be used to update picture on demand

phattmatt:
I personally like the all-in-one nature of InkyPi.

Not having a dependency on an external server/service suits my needs better than needing to have something else running elsewhere on my network.

It makes the frame more portable as well. It will run anywhere with access to the Internet.

That said, making the service that manages the image to be displayed (playlist, plugins, API access, scheduling, etc), and a service that shows the image on the display (takes current image and sends to the display) separate, but runnable on the same device, or different devices might be possible.

I suspect this would take quite a lot of effort to reengineer InkyPi, so would need someone  to dedicate a lot of time to achieving this.

dJason:
Next step following the merge of #402 above would be to allow for a "Dummy" display installation (similar to the --dev mode), where you configure the dimensions and properties of the display post-install.
This would also be a stepping stone towards IT8951 support, using the compiled binary driver to push the rendered image to the display rather than using the Python EPDs.

PR #402, opened 2025-11-11
Add support for current_image polling

deftdawg:
- add support for polling via "If-Modified-Since" request header / HTTP 304 response.
- make admin UI poll every 3 seconds for current_image updates, if no update no content returned (HTTP 304).
- polling support also enables dumb frame devices (i.e. [Neoframe](https://github.com/deftdawg/neoframe)/[ESP32](https://www.good-display.com/product/574.html)) to stay updated by simply keeping track of last HTTP 200 date/time.

### DIFF
--- a/src/blueprints/main.py
+++ b/src/blueprints/main.py
@@ -1,4 +1,6 @@
-from flask import Blueprint, request, jsonify, current_app, render_template
+from flask import Blueprint, request, jsonify, current_app, render_template, send_file
+import os
+from datetime import datetime
 
 main_bp = Blueprint("main", __name__)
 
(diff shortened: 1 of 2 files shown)


fatihak:
Thanks for the PR! This is a solid improvement and adds support for building low-power clients like ESP32-based frames that periodically wake up, poll, and go back to sleep. That’s been one of the most requested features (#215), so this really helps move things forward.

A natural next step for a client/server model would be the ability to run an InkyPi installation without an attached display and let it act purely as an always-on “server.” From there, we could provide example ESP32 polling scripts and some documentation so people can easily pair low-power remote displays with a central InkyPi instance.

PR #405, opened 2025-11-13
Dark Mode

deftdawg:
- Adds dark mode throughout the UI
- Sets open-meteo as default weather provider and hides API key label unless OpenWeatherMap is selected


deftdawg:
> Only issue i'm having is that all the pages seem to be loading in light mode by default

I don't see that on my system (zen browser/dark reader/nixos/KDE dark theme)... 

I'll have to test with dark reader off, but I think my browser opens new tabs with dark background from the system theme.  Is your system running a dark mode theme?

fatihak:
Attaching a video here. My system is on light mode, but it might be due to my network requests taking longer? Not sure
https://github.com/user-attachments/assets/2b9a497f-f079-4f3b-a82b-e0fa575d5ba6

deftdawg:
I’ll try to reproduce it w/ chromium and fix it

deftdawg:
Reproduced on chromium and fixed, should be good everywhere :crossed_fingers:  

EDIT: shoot, conflicts... gotta fix 'em

fatihak (approved):
Looks good!

PR #410, opened 2025-11-14
Add HEIF/HEIC image support

roberthunt:
This adds support for HEIF/HEIC images to the image upload, image folder and image URL plugins. Also for any plugin that might provide images via `handle_request_files`.

Uses `pi-heif` instead of `pillow-heif` as only image HEIF decoding is required and not saving / writing.

### DIFF
--- a/src/inkypi.py
+++ b/src/inkypi.py
@@ -2,6 +2,9 @@
 
 # set up logging
 import os, logging.config
+
+from pi_heif import register_heif_opener
+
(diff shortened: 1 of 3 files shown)


fatihak:
Hm the image upload plugin saves images uploaded to the form, does this work if pi-heif doesn't allow saving/writing?

fatihak:
Nevermind I was able to test and see it working, thanks!

fatihak (approved):

PR #414, opened 2025-11-15
Add auto-setup/auto-build via jetify devbox on MacOS and Linux 

deftdawg:
This is a PR to use devbox to make a reproducible build toolchain that auto-installs/sets-up everything needed to run on any Linux (Ubuntu/Fedora/Arch/NixOS) or MacOS system.

To install devbox: `curl -fsSL https://get.jetify.com/devbox | bash`
To activate devbox: `devbox shell` (from within the project directory)
or just `devbox run dev` to run developer version of InkyPi on 8080

- add devbox+direnv config to auto-install build tools,  a python virtual environment and display a dev welcome message via "devbox shell"
- update venv.sh to use bash in PATH and accept external VENV_DIR variable locations, and resolve SRC_DIR to a full path (fixing render file /tmp refering fonts via a relative path)
- update image_utils to automatically pick the first availble chromium-like browser in the path ("chromium" is used on NixOS/Fedora/etc where "chromium-headless-shell" isn't available)
- Update update_vendors.sh to use bash from PATH, pull down Full calendar CSS instead of JS twice (bugfix), externalize versions into variables
- Update devbox.json to run update_vendors.sh if calendar or select2 files are missing


deftdawg:
@fatihak easy dev setups on MacOS, Linux and WSL2 ready to merge

fatihak:
Can you instructions to setup/run devbox to https://github.com/fatihak/InkyPi/blob/main/docs/development.md
We should keep existing instructions to run locally there, but have this listed as an alternative method.

deftdawg:
Updated the `development.md`  doc, I don't have a Windows machine, so I can't test the Windows instructions, but I've used devbox w/ Chromium before within WSL2 for other stuff on Windows and it  worked.

fatihak (approved):
Looks good!

Issue #426, opened 2025-11-20
Unsplash crash with no helpful errors


iAndrea:
**Describe the bug**
Using Unsplash to grab random images makes the rasp crash after a while with no helpful error. My guess is that it is not able to process big resolutions, but I see no error that can prove it. Did someone else experienced it?

I kept `htop` open during refreshed and I see normal behavior, no RAM or SWAP filled, just a sudden crash.

Example of images that made it crash:

- https://images.unsplash.com/photo-1546525848-3ce03ca516f6?crop=entropy&cs=srgb&fm=jpg&ixid=M3w4MzI2MTl8MHwxfHNlYXJjaHwzfHxibGFjayUyMG1hbnxlbnwwfDB8fHwxNzYzNTg3MzUxfDA&ixlib=rb-4.1.0&q=85
- https://images.unsplash.com/photo-1545198311-96dcf151faa4?crop=entropy&cs=srgb&fm=jpg&ixid=M3w4MzI2MTl8MHwxfHNlYXJjaHw5fHxjaHJpc3RtYXMlMjBkZWNvcnxlbnwwfDB8fHwxNzYzNTQ5NDc1fDA&ixlib=rb-4.1.0&q=85

**Expected behavior**
Being able to use Unsplash without crashes every x images.

**To Reproduce**
1. Set Unsplash plugin to update image, or force a bunch un reloads
2. Wait for the  inky service to crash

**Logs**
```
inkypi[5361]: /usr/local/bin/inkypi: line 13:  5365 Killed                  python -u "$(realpath $PROGRAM_PATH/src/inkypi.py)"
systemd[1]: inkypi.service: Deactivated successfully.
systemd[1]: inkypi.service: Consumed 33.684s CPU time.
```

**Additional context**
Raspberry Pi Zero 2W, Inky Impression 7", Raspbian Full

iAndrea:
I see in the code we are fetching with:
```
image_url = random.choice(results)["urls"]["full"]
```

But instead we could provide in the UI the choices that Unsplash search API provides:
```
"urls": {
        "raw": "https://images.unsplash.com/photo-1416339306562-f3d12fefd36f",
        "full": "https://hd.unsplash.com/photo-1416339306562-f3d12fefd36f",
[...]

PR #427, opened 2025-11-20
Image memory optimization to prevent OOM crashing and other improvements

iAndrea:
Following Issue: https://github.com/fatihak/InkyPi/issues/426

### DIFF
--- a/install/install.sh
+++ b/install/install.sh
@@ -309,8 +309,8 @@ install_src() {
 }
 
 install_cli() {
-  cp -r "$SCRIPT_DIR/install/cli" "$INSTALL_PATH/"
-  sudo chmod +x "$INSTALL_PATH/cli/"*.sh
+  cp -r "$SCRIPT_DIR/cli" "$INSTALL_PATH/"
+  sudo chmod +x "$INSTALL_PATH/cli/"*
 }
 
 # Get Raspberry Pi hostname
--- a/src/blueprints/playlist.py
+++ b/src/blueprints/playlist.py
@@ -72,11 +72,13 @@ def playlists():
     device_config = current_app.config['DEVICE_CONFIG']
     playlist_manager = device_config.get_playlist_manager()
     refresh_info = device_config.get_refresh_info()
+    plugins_list = device_config.get_plugins()
 
     return render_template(
         'playlist.html',
         playlist_config=playlist_manager.to_dict(),
-        refresh_info=refresh_info.to_dict()
+        refresh_info=refresh_info.to_dict(),
+        plugins={p["id"]: p for p in plugins_list}
     )
 
 @playlist_bp.route('/create_playlist', methods=['POST'])
--- a/src/display/waveshare_display.py
+++ b/src/display/waveshare_display.py
@@ -83,7 +83,6 @@ class WaveshareDisplay(AbstractDisplay):
             self.epd_display_init()
 
             display_args_spec = inspect.getfullargspec(self.epd_display.display)
-            display_args = display_args_spec.args
         except ModuleNotFoundError:
             raise ValueError(f"Unsupported Waveshare display type: {display_type}")
         except AttributeError:
--- a/src/plugins/ai_image/ai_image.py
+++ b/src/plugins/ai_image/ai_image.py
@@ -115,7 +115,7 @@ class AIImage(BasePlugin):
 
     @staticmethod
     def fetch_image_prompt(ai_client, from_prompt=None):
-        logger.info(f"Getting random image prompt...")
+        logger.info("Getting random image prompt...")
 
         system_content = (
             "You are a creative assistant generating extremely random and unique image prompts. "
--- a/src/plugins/calendar/calendar.py
+++ b/src/plugins/calendar/calendar.py
@@ -49,7 +49,7 @@ class Calendar(BasePlugin):
         logger.debug(f"Fetching events for {start} --> [{current_dt}] --> {end}")
         events = self.fetch_ics_events(calendar_urls, calendar_colors, tz, start, end)
         if not events:
-            logger.warn("No events found for ics url")
+            logger.warning("No events found for ics url")
 
         if view == 'timeGridWeek' and settings.get("displayPreviousDays") != "true":
             view = 'timeGrid'
--- a/src/plugins/github/github_contributions.py
+++ b/src/plugins/github/github_contributions.py
@@ -70,7 +70,7 @@ def fetch_contributions(username, api_key):
 def parse_contributions(data, colors):
     weeks = data["data"]["user"]["contributionsCollection"]["contributionCalendar"]["weeks"]
 
-    grid = [[day for day in week["contributionDays"]] for week in weeks]
+    grid = [list(week["contributionDays"]) for week in weeks]
     max_contrib = max(day["contributionCount"] for week in grid for day in week)
 
     def get_color(count):
--- a/src/plugins/github/github_stars.py
+++ b/src/plugins/github/github_stars.py
@@ -35,7 +35,6 @@ def stars_generate_image(plugin_instance, settings, device_config):
     )
 
 def fetch_stars(github_repository):
-    global data
     url = f"https://api.github.com/repos/{github_repository}"
     headers = {"Accept": "application/json"}
 
@@ -44,6 +43,7 @@ def fetch_stars(github_repository):
         data = response.json()
     else:
         logger.error(f"GitHub Stars Plugin: Error: {response.status_code} - {response.text}")
+        data = {"stargazers_count": 0}
 
     return data['stargazers_count']
 
--- a/src/plugins/image_folder/image_folder.py
+++ b/src/plugins/image_folder/image_folder.py
@@ -79,6 +79,7 @@ class ImageFolder(BasePlugin):
                 logger.debug(f"Scaling to fit dimensions: {dimensions[0]}x{dimensions[1]}")
                 img = ImageOps.fit(img, dimensions, method=Image.LANCZOS)
 
+            return img
         except Exception as e:
             logger.error(f"Error loading image from {image_url}: {e}")
             raise RuntimeError("Failed to load image, please check logs.")
--- a/src/plugins/newspaper/newspaper.py
+++ b/src/plugins/newspaper/newspaper.py
@@ -27,7 +27,7 @@ class Newspaper(BasePlugin):
             image_url = FREEDOM_FORUM_URL.format(date.day, newspaper_slug)
             image = get_image(image_url)
             if image:
-                logging.info(f"Found {newspaper_slug} front cover for {date.strftime('%Y-%m-%d')}")
+                logger.info(f"Found {newspaper_slug} front cover for {date.strftime('%Y-%m-%d')}")
                 break
 
         if image:
--- a/src/plugins/plugin_registry.py
+++ b/src/plugins/plugin_registry.py
@@ -15,17 +15,17 @@ def load_plugins(plugins_config):
     for plugin in plugins_config:
         plugin_id = plugin.get('id')
         if plugin.get("disabled", False):
-            logging.info(f"Plugin {plugin_id} is disabled, skipping.")
+            logger.info(f"Plugin {plugin_id} is disabled, skipping.")
             continue
 
         plugin_dir = plugins_module_path / plugin_id
         if not plugin_dir.is_dir():
-            logging.error(f"Could not find plugin directory {plugin_dir} for '{plugin_id}', skipping.")
+            logger.error(f"Could not find plugin directory {plugin_dir} for '{plugin_id}', skipping.")
             continue
 
         module_path = plugin_dir / f"{plugin_id}.py"
         if not module_path.is_file():
-            logging.error(f"Could not find module path {module_path} for '{plugin_id}', skipping.")
+            logger.error(f"Could not find module path {module_path} for '{plugin_id}', skipping.")
             continue
 
         module_name = f"plugins.{plugin_id}.{plugin_id}"
--- a/src/plugins/weather/render/weather.html
+++ b/src/plugins/weather/render/weather.html
@@ -243,8 +243,8 @@
               },
               autoSkip: false,
               callback: function(value, index, values) {
-                if (index === values.length-1) return maxTemp + "°";
-                else if (index === 0) return minTemp + "°";
+                if (index === values.length-1) return (units != "standard") ? maxTemp + "°" : maxTemp;
+                else if (index === 0) return (units != "standard") ? minTemp + "°" : minTemp;
                 else return '';
               }
             },
--- a/src/plugins/weather/settings.html
+++ b/src/plugins/weather/settings.html
@@ -245,8 +245,8 @@
             document.getElementById('displayGraph').value = "true";
 			document.getElementById('displayRain').checked = false;
             document.getElementById('displayRain').value = "false";
-			document.getElementById('displayGraphIcons').checked = true;
-			document.getElementById('displayGraphIcons').value = "true";
+			document.getElementById('displayGraphIcons').checked = false;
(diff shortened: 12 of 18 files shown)


deftdawg:
I don’t have unsplash, but I do have 13.3” display (resolution is 1600x1200).   Are all unsplash images standard sizes based on what’s selected?  I’m just wondering if you can auto select the appropriate size based on the resolution in the device settings?

iAndrea:
According to [Unsplash documentation](https://unsplash.com/documentation#search) there are 5 "sizes" for their images:
- raw
- full (what is used currently in the main branch)
- regular, meaning images with width of 1080px
- small, [...] width of 400px
- thumb, [...] with of 200px
I gave the possibility to choose between full, regular and small with a new dropdown in the UI.

I think it wouldn't be the best solution to auto-choose a specific format based on the display in use because being a memory related problem the choice is more about which raspberry we are using rather than the display itself.
Does it make sense? @deftdawg

iAndrea:
New commit updates accordion to https://github.com/fatihak/InkyPi/issues/426#issuecomment-3567573878

**Changed Default Image Size**

Changed from requesting "full" (6000px) to "regular" (default - 1080px) from Unsplash API.
Simple but effective for preventing crashes

**Added User-Selectable Image Quality (Frontend)**

Created a dropdown in settings.html with 3 options:
- Small (400px) - Fastest, lowest quality
- Regular (1080px) - Recommended, perfect for e-ink
- Full (2000-4000px) - High quality, risky on Pi Zero
Each option saves with the plugin instance in playlists.

**Automatic Safety Protection (Backend)**

Added device detection in unsplash.py:
```
# Detects if device has < 1GB RAM
if is_low_resource and image_size == 'full':
      image_size = 'regular'  # Auto-downgrade
      logger.warning("Automatically downgraded for stability")
```

- Pi Zero automatically protected from 'full' size
- Logs clear warnings when downgrade happens
- More powerful devices (Pi 3/4) can use 'full' without restrictions

**Memory Processing Optimizations**
Improved the grab_image() function:
- Two-stage resize for large images (downsample → high-quality resize)
- In-place thumbnail() instead of copy()
- RGB conversion to remove alpha channels
- Adaptive quality filters: BICUBIC on Pi Zero, LANCZOS on powerful devices
- Explicit garbage collection after processing stages

**Comprehensive Logging**
Added detailed logging at multiple levels:
- DEBUG: Every step (RAM detection, resize strategy, dimensions, etc.)
- INFO: Key events (image downloaded, processing complete)
- WARNING: Automatic downgrades and safety measures
- ERROR: Failures with helpful suggestions
Can be set in `src/config/logging.conf`

deftdawg:
This is the way! :smile: 

Some other things I was considering tinkering with:
- Saving image to a temp file and streaming it to resize, rather than downloading it in mem and then resizing it
- Centralizing download and re-scaling of images into the parent "plugin" class, since this is an issue that is likely to occur for unsplash, wpotd, apod, image_url, image_folder, image_upload, ai_image(maybe)...

iAndrea:
@deftdawg Got it!

- temp file -> I'm already working on it 🎉 
- centralization -> I'll need more time but sure, I already though about it since it is definitely a common problem throughout the project since low ram is low ram everywhere 🙃

iAndrea:
**Latest Changes**

I went for a dual-path process since writing in a temp file is quite a bit slower than RAM due to sd card performances. Still is the best solution for low ram devices, whilst for standard devices I opted to keep the process in-memory to speed things up a bit... so:
- Implemented dual-path image loading:
  - Low-resource path: Stream to temp file + PIL draft mode
  - High-performance path: In-memory processing (unchanged)
- Added PIL draft() mode to decode images at reduced resolution (80-90% memory savings)
- Implemented chunk-based streaming (8KB chunks) to minimize download memory
- Added automatic temp file cleanup in finally block
- Enhanced logging to show which processing path is used

The PIL's draft mode is a game changer, even with RAW images I have no crashes in my Zero 2W 🚀 
Try it out and let me know :)
Do you want to merge this Unsplash-related branch and solve it, or wait for the centralization? Again, for the centralization I'll need time but I'll do it. This is project is awesome!

deftdawg:
Excellent work, I wish I could test or merge it… but hopefully @fatihak will get a chance to look at it soonish.  

Below is rough patch cooking for wpotd, which I was looking into …

  - max resolution set to 2K (I think 28.5 e-ink are max 2K, so nothing larger makes sense)
  - Not yet orientation aware (vertical vs horizontal)
  - for png seems reduce(), for jpeg draft
  - All completely untested 


~~~diff
diff --git a/src/plugins/wpotd/wpotd.py b/src/plugins/wpotd/wpotd.py
index 35d3dc4..ab3dcf0 100644
--- a/src/plugins/wpotd/wpotd.py
+++ b/src/plugins/wpotd/wpotd.py
@@ -26,10 +26,13 @@ from PIL import Image, UnidentifiedImageError
 from io import BytesIO
 import requests
 import logging
+import os
+import tempfile
 from random import randint
 from datetime import datetime, timedelta, date
 from functools import lru_cache
 from typing import Dict, Any
+import math
 
 logger = logging.getLogger(__name__)
 
@@ -79,9 +82,46 @@ class Wpotd(BasePlugin):
                 logger.warning("SVG format is not supported by Pillow. Skipping image download.")
                 raise RuntimeError("Unsupported image format: SVG.")
 
-            response = self.SESSION.get(url, headers=self.HEADERS, timeout=10)
-            response.raise_for_status()
-            return Image.open(BytesIO(response.content))
+            # Stream to a temporary file to avoid loading the whole compressed image into RAM
+            fd, temp_path = tempfile.mkstemp()
+            os.close(fd)
+            
+            try:
+                with self.SESSION.get(url, headers=self.HEADERS, stream=True, timeout=30) as response:
+                    response.raise_for_status()
+                    with open(temp_path, 'wb') as f:
+                        for chunk in response.iter_content(chunk_size=8192):
+                            f.write(chunk)
+                
+                with Image.open(temp_path) as img:
+                    # Optimize for memory: 
+                    # The user requested to fit within 2K resolution (2560x1600)
+                    target_width, target_height = 2560, 1600
+                    
+                    if img.format == 'JPEG':
+                        img.draft('RGB', (target_width, target_height))
+                    elif hasattr(img, 'reduce'):
+                        # For non-JPEGs (like PNG), use reduce if available
+                        width, height = img.size
+                        factor_w = width / target_width
+                        factor_h = height / target_height
+                        factor = max(factor_w, factor_h)
+                        
+                        if factor > 1:
+                            reduce_factor = math.ceil(factor)
+                            # reduce returns a new image
+                            img = img.reduce(reduce_factor)
+                            
+                    img.load()
+                    # Return a copy to ensure independence from the closed file
+                    # If img is a new object from reduce(), copy() is fine but maybe redundant if we just return img.
+                    # But if it's the original img object (JPEG case), we need copy() because of the file close.
+                    # To be safe, always copy.
+                    return img.copy()
+            finally:
+                if os.path.exists(temp_path):
+                    os.remove(temp_path)
+
         except UnidentifiedImageError as e:
             logger.error(f"Unsupported image format at {url}: {str(e)}")
             raise RuntimeError("Unsupported image format.")
~~~

Besides the bullets above, centralized download_image could also consider:

 - local files won’t need temp file for processing (url starts with “/“ or “file://“)
 - provide optional parameters for device resolution and orientation (if multi-plugin layouts become a thing, this could be useful) — default to device values for now

iAndrea:
@deftdawg I'm pretty close to completion, just a bunch of tests and I'll push the changes for a central image management :thumbnail

iAndrea:
So here we are with the next update: "Centralize image processing with adaptive memory optimization"

**Problem**
Multiple plugins were independently downloading and processing images with no memory optimizations, causing potential crashes on Raspberry Pi Zero 2W (512MB RAM) when handling high-resolution images (4000x6000+). Each plugin had duplicate image loading code with inconsistent error handling.

**Solution**
Implemented centralized _AdaptiveImageLoader_ utility that automatically optimizes based on device capabilities:

1. Low-resource devices (Pi Zero, <1GB RAM):
 - Streams downloads to temp files (8KB chunks, not full image in RAM)
 - Uses PIL draft mode for 80-90% memory reduction during JPEG decode
 - Two-stage resize (thumbnail → LANCZOS) for huge images
 - Explicit garbage collection after each stage
2. High-performance devices (Pi 3/4, ≥1GB RAM):
 - Fast in-memory processing
 - Full-resolution decode with LANCZOS (highest quality)
 - No disk I/O overhead

The new utility `src/utils/image_loader.py` is integrated into BasePlugin so then:
 - All plugins now have self.image_loader available
 - Simple API: .from_url(), .from_file(), .from_bytesio()

Migrated 8 plugins:
✅ apod
✅ image_url
✅ wpotd
✅ image_folder
✅ image_upload
✅ ai_image
✅ comic
✅ unsplash
Each has been tested tested on Pi Zero 2W except for apod, image_folder, ai_image.

Logging:
 - Added comprehensive INFO/DEBUG/ERROR logging to all migrated plugins
 - Consistent format matching Unsplash implementation
 - Shows device detection, processing strategy, and optimization steps

iAndrea:
Also a small side-note that I forgot.

- Removed Dead Code (__shrink_to_fit()_ method) in wpotd.py
  - This is now completely redundant since AdaptiveImageLoader handles all resizing
- Implemented Connection Pooling
  - Created src/utils/http_client.py handling shared requests.Session() singleton
  - Connection pooling (max 10 connections per host)
  - Automatic retry on failure (3 retries)
  - Keep-alive enabled by default
  - Consistent User-Agent header across all requests

Updated 6 plugins + image_loader to use shared session:
✅ apod.py
✅ wpotd.py
✅ unsplash.py
✅ ai_image.py
✅ comic.py
✅ image_loader.py
NB: same tested / tested I said earlier, so **there are 3 plugins that need testing**

---

There are still a couple of minor changes that I was considering but I would like your opinion :)
1. APOD HD Quality Toggle
- Add user setting: "Use HD quality" (default: false)
- HD URLs are 10MB+ files, standard is often enough
2. Reduce Unsplash per_page
- Change from 100 to 30 results
- Smaller API response, still plenty of variety (I guess)

deftdawg:
Nice work, reading through it the code looks good to me.  You should probably rename the PR title since it's a little more than unsplash quality selection now. 😅

iAndrea:
@deftdawg Done. Anyway I took inspiration from what you posted about your wpotd changes, so thank you :)

iAndrea:
@fatihak Can you take a look into this PR? I think it would improve quite a lot the project stability for low-memory devices 🙏

iAndrea:
Due to the introduction of the new `image_album` plugin with support for Immich, this new plugin is subject to the same problems we already talked about. In fact I tried to retrieve a picture of mine with resolution 6000x4000 and it crashed. I've already updated that plugin to support the new features:
- Replaced Direct HTTP Requests with Centralized Session
- Integrated AdaptiveImageLoader
- Enhanced Logging
  - Added comprehensive logging throughout the plugin (matching patterns from unsplash, apod, image_url)
  - Detailed logging for configuration, dimensions, and processing steps
  - Better error messages with context
- Improved Error Handling
  - Fixed bug in get_album_id() where it would fail if album not found
  - Added check for empty asset lists
  - Better exception messages with more context
- Match Orientation awareness
  - Selects only images matching the device settings according to exif (dimensions + orientation flags)
  - For images with slightly different ratio compared to the device, using crop-to-fill instead of pad-to-fit for optimal display

deftdawg:
Not to minimize the other excellent work done here, but I would rename this PR ticket to `Image memory optimization to prevent OOM crashing and other improvements`, I think keeping InkyPi from dying from OOM when it loads big images is the most important aspect from a community perspective - most of the problems people currently report with Inky dying are very likely caused by OOM that this resolves...   @fatihak

iAndrea:
@fatihak Is there a reason why we cannot merge this request as a Christmas gift for every rpi zero 2 user? Do you need me to add something? :)

src/plugins/wpotd/wpotd.py
```diff
@@ -44,26 +43,42 @@ def generate_settings_template(self) -> Dict[str, Any]:
...
+        # Get dimensions
+        max_width, max_height = device_config.get_resolution()
+        if device_config.get_config("orientation") == "vertical":
+            max_width, max_height = max_height, max_width
+            logger.debug(f"Vertical orientation detected, dimensions: {max_width}x{max_height}")
 
+        dimensions = (max_width, max_height)
+
+        # Use adaptive loader if shrink-to-fit is enabled
+        if settings.get("shrinkToFitWpotd") == "true":
```
fatihak: This can just be simplified to self._download_image(picurl, dimensions=dimensions, resize= settings.get("shrinkToFitWpotd") == "true")

src/plugins/image_album/settings.html
```diff
@@ -21,6 +21,25 @@
...
+        <label for="orientationFilter" class="form-label">Match Orientation:</label>
+        <div class="toggle-container">
+            <!-- Checkbox for display only, doesn't submit -->
+            <input type="checkbox" id="orientationFilterCheckbox" class="toggle-checkbox" checked
+                onchange="updateOrientationFilterValue()">
+            <label for="orientationFilterCheckbox" class="toggle-label"></label>
+        </div>
+        <!-- Hidden input that actually submits the value -->
+        <input type="hidden" id="orientationFilter" name="orientationFilter" value="true">
+        <small style="display: block; margin-top: 5px; color: #666; font-size: 0.9em;">
```
fatihak: This feature isn't needed and isn't related to the memory optimization

src/plugins/image_album/image_album.py
```diff
@@ -1,74 +1,239 @@
...
+            if 'exifInfo' in first_asset:
+                exif_keys = list(first_asset['exifInfo'].keys())
+                logger.debug(f"EXIF keys: {exif_keys}")
+                exif_info = first_asset['exifInfo']
+                if 'exifImageWidth' in exif_info and 'exifImageHeight' in exif_info:
+                    logger.debug(f"Dimensions: {exif_info['exifImageWidth']}x{exif_info['exifImageHeight']}")
+
+        return detailed_assets
+
+    def filter_assets_by_orientation(self, assets: list[dict], dimensions: tuple[int, int]) -> list[dict]:
```
fatihak: Not needed

src/plugins/unsplash/settings.html
```diff
@@ -45,17 +45,32 @@
             <option value="portrait">Portrait</option>
         </select>
     </div>
+
+    <div class="form-group">
```
fatihak: Not sure why this is needed if the code forces downsizing on low-resource devices. I think we should go with regular size on low-resource devices and full otherwise, without this option in the UI.

fatihak:
Thanks for the changes. Also just realized when testing that the behavior of the "Scale to Fit" option is changed. We don't want to stretch images with this option, it should just scale the image, keeping the aspect ratio, to fit the dimensions of the screen. 
Here is a sample image I used:
test_image

Image displayed previously with 'Scale to Fit" option enabled:
IMG_1608


Image displayed now with 'Scale to Fit' option enabled:
IMG_1609

src/plugins/image_album/settings.html
```diff
@@ -21,6 +21,9 @@
         </div>
         <!-- Hidden input ensures 'true' is always sent when unchecked -->
         <input type="hidden" name="padImage" value="true">
+        <small style="display: block; margin-top: 5px; color: #666; font-size: 0.9em;">
```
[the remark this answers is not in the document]
iAndrea: Sure 👍

iAndrea:
@fatihak Got it, now using the fit, should be fixed

fatihak:
I saw this issue in the image upload plugin, but seems like you only updated the image album plugin

iAndrea:
@fatihak I managed to fix it directly into the AdaptiveImageLoader utility 👍

PR #430, opened 2025-11-21
Add selecting headless chromium from candidates

efrecon:
When developping, `chromium-headless-shell` is not available in all Linux distros. In addition, on Windows and macOS, development instructions mentioned locally modifying the code. This PR replaces #400 with a simpler solution. It:

+ Picks the chrome-derivative binary to use for rendering from the `PATH`.
+ Adapts documentation to the new behavior.

Note: I have not tested all binaries. The list is a mix of binaries that I know of, binaries that were documented before (Use of chrome headless on Windows and macOS) and "good guesses". I expect the community to tweak and amend the list with time.


fatihak:
Closing this in favor of https://github.com/fatihak/InkyPi/pull/414. I'm not sure about adding untested binaries, that PR added ["chromium-headless-shell", "chromium", "chrome"]. I would prefer the community add binaries that are verified to work over time.

PR #435, opened 2025-11-24
Add dynamic hostname setting based on MAC address

dsbilling:
Introduced functions to retrieve the MAC address and set the system hostname to the app name plus the last 4 characters of the MAC address. This ensures unique hostnames for each device during installation.

This ensures each device has a unique hostname during installation, preventing network conflicts and making devices easily identifiable.

### DIFF
--- a/install/update.sh
+++ b/install/update.sh
@@ -48,12 +48,65 @@ get_os_version() {
   echo "$(lsb_release -sr)"
 }
 
+# Get MAC address from primary network interface
+get_mac_address() {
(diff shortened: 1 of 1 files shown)


dsbilling:
@fatihak Adding the "set hostname" in the `update.sh` file might be too destructive. What do you think?

fatihak:
Wondering if it might be more straightforward to have an optional argument to the install script to set a custom hostname instead?

fatihak:
Haven't heard back, I also think updating the host name after an update would be destructive and confusing. I would be open to having an argument in the install/update scripts to be able to set a custom hostname instead.

PR #441, opened 2025-11-28
Adding plugin: upcoming valorant champion tour match displayer

Abudidayo:
Using a 3rd party rest api to get games info and displays the latest game. 

This is my first PR so im not sure if this is how its supposed to me made.


fatihak:
Thanks for the PR, this seems like a cool plugin. That said, it’s a bit too specific to a single game/event to include in the core project. I’m trying to keep the built-in plugins fairly general-purpose so they’re useful to most users.

PR #451, opened 2025-12-06
Run Once mode & Render error messages on frame

deftdawg:
- "Run next plugin and exit" mode (--run-once)
- A simple error message now renders on the frame when a plugin fails


deftdawg:
Simple error messages look like this:
(changed comic plugin font to "Tommy" to break it)

image

PR #454, opened 2025-12-08
Add dynamic/custom hostname during installation

dsbilling:
Reworked from #435


dsbilling:
@fatihak Sorry for delayed response, here is the rework 👍🏻

PR #457, opened 2025-12-09
Orphan images cleanup & plugin thumbnails preview

iAndrea:
Added automatic image cleanup for playlist / plugin removal and thumbnail preview for each plugin instance inside a playlist.
Screenshot 2025-12-09 at 17 26 58
Screenshot 2025-12-09 at 17 27 02

During the image deletion these logs appear:
```
INFO - blueprints.plugin - Deleted plugin instance image: src/static/images/plugins/clock_clock3.png
```

### DIFF
--- a/src/blueprints/plugin.py
+++ b/src/blueprints/plugin.py
@@ -9,43 +9,10 @@ import logging
 logger = logging.getLogger(__name__)
 plugin_bp = Blueprint("plugin", __name__)
 
-def _delete_uploaded_images_from_settings(settings, deleted_paths=None):
-    """Recursively find and delete uploaded images from plugin settings."""
-    if deleted_paths is None:
-        deleted_paths = set()
-
-    saved_images_dir = resolve_path(os.path.join("static", "images", "saved"))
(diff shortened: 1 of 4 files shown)


fatihak:
Thanks for the PR, i like the thumbnail feature and the deletion of files. In addition to the comment above, my only other feedback is that if a plugin doesn't have a thumbnail it either shouldn't have a preview or we should have a placeholder preview image. This is how it currently looks on chrome:

Screenshot 2025-12-10 at 8 58 52 AM

iAndrea:
@fatihak Sure I was already working on the broken preview. It was working initially, I don't know what I've done 🙃  
Moved the cleanup method into the base_plugin for letting single plugins handle its own process 👍 
Screenshot 2025-12-10 at 15 45 05

iAndrea:
@fatihak You're welcome, I love this project :)
There is also this https://github.com/fatihak/InkyPi/pull/427 that will help a lot of users with their pi 2 zero's like mine 🙏

PR #458, opened 2025-12-10
Add wind direction

k2v1n58:
In this PR, I add arrows corresponding to the wind direction after the wind speed
For example:

- ↑ if the wind direction is between $337.5^\circ \text{ and} 22.5^\circ$
- ↙ if the wind direction is between  $202.5^\circ \text{ and } 247.5^\circ$

image


fatihak:
Looks good! Thanks for the PR!

PR #472, opened 2025-12-21
Fix Open-Meteo timezone handling for sunrise/sunset/hourly times

markbreen:
Problem
  Open-Meteo API returns times already in local timezone when timezone=auto is specified. Calling .astimezone() incorrectly re-converted these times, causing them to display wrong when the system timezone differed from the weather location.

  Changes
  - Removed incorrect .astimezone(tz) calls from sunrise, sunset, and hourly time parsing


fatihak:
Hi, the existing code is converting the timezone to the timezone that is configured for your device from the settings page. I just tested with my display and I'm seeing the correct times. You might need to update the timezone for your display if it's not displaying the right times.

PR #482, opened 2025-12-25
Use miles for weather visibility when units are set to imperial

### LINKED ISSUE #239: Weather Plugin - Visability Issue - Not Respecting Saved Units of Measure

jjacobowitz:
Fixes #239 by switching to miles for the weather visibility when the units system is set to imperial. Changes were made for both OpenWeather (previously used km regardless of unit system) and open-meteo (previously used ft when set to imperial). I've also swapped the `>` symbol for `u"\u2265"` (`≥`) because the visibility will be greater than _or equal to_ the listed visibility.

Note: OpenWeather has a maximum visibility of 10 km (6.2 miles) while open-meteo does not appear to have a maximum. To be consistent, should the visibility from open-meteo be capped at the same 10 km (6.2 miles)?

### DIFF
--- a/src/plugins/weather/weather.py
+++ b/src/plugins/weather/weather.py
@@ -32,16 +32,19 @@ def get_moon_phase_name(phase_age: float) -> str:
 UNITS = {
     "standard": {
         "temperature": "K",
-        "speed": "m/s"
+        "speed": "m/s",
+        "distance":"km"
     },
     "metric": {
         "temperature": "°C",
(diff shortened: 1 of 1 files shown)


fatihak:
Looks good, thanks for the PR! And i think the max visibility logic you have is fine :)

PR #489, opened 2025-12-27
Suggestion: allow save another instance when editing

GuilhermeMorais:
This pull request introduces a new "Save As..." button to the plugin details page when editing an existing plugin instance. This provides users with an additional option to save the current plugin instance under a new configuration or name.

image

User interface enhancements:

* Added a "Save As..." button (which opens the `scheduleModal`) next to the existing "Save" button for plugin instances in the `plugin.html` template.


fatihak (approved):
Thanks for the PR, didn't think this would be this simple but makes sense. Merged.

PR #503, opened 2026-01-03
Allow plugins to have different screen time (custom plugin_cycle_interval_seconds)

rs9000:
This pull request try to adds support for per-plugin display duration control by introducing a **sleep_time** attribute in plugin configuration file (**plugin-info.json**). 

Previously, all plugins shared the same display time interval defined by the global **plugin_cycle_interval_seconds** setting. Now, each plugin can specify its own custom sleep time, allowing for more flexible playlist configurations.

Example:

Playlist1
- Image Album Plugin: 30m
- Weather Plugin: 15m

Backward compatibility:
Plugins without a sleep_time attribute continue to use the global default setting (plugin_cycle_interval_seconds)

### DIFF
--- a/src/refresh_task.py
+++ b/src/refresh_task.py
@@ -45,6 +45,25 @@ class RefreshTask:
         if self.thread:
             logger.info("Stopping refresh task")
             self.thread.join()
+            
+    def _get_sleep_time(self):
+        """Gets the sleep time from the last refreshed plugin config, or falls back to device default."""
+        default_sleep_time = self.device_config.get_config("plugin_cycle_interval_seconds", default=60*60)
+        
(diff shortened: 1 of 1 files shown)


fatihak:
Thanks for the PR and the detailed explanation, I understand the motivation behind per-plugin screen timing.

However, the existing refresh and scheduling settings are already confusing for some users and adding another layer of timing configuration would increase that complexity. I don’t think this is necessary for most use cases.

The current approach also aligns with how TRMNL handles this, using a single standard “Refresh Rate,” which keeps things simpler.

For these reasons, I’m not planning to move forward with this change. Thanks again for the contribution.

Issue #502, opened 2026-01-03
[Feature Request] Change saturation default, or add configurability, for Inky Impression driver


phattmatt:
As described in this issue:

[Poor display (possible scaling problem) with Inky Impression 7.3 2025 Edition PIM773](https://github.com/fatihak/InkyPi/issues/111)

The default saturation for the Inky Impression driver may not result in the best display quality for many plugins (due to the image type).

A link to the description of why the driver works in this manner, and suggestions of what to do about it in the context of InkyPi can be found here: https://github.com/fatihak/InkyPi/issues/111#issuecomment-3215466577

> Looks like the "saturation" parameter behaviour is expected; it's not really "saturation", but a method of quantizing the palette values before dithering to fine tune the result for different images types. 
> 
> There are a few options available to us:
> 
> 0) Do nothing, leave the default as '0.5'.
> 1) Statically set the Inky 'saturation' value to be '0', since (IMHO), this is more appropriate for the majority of images shown by InkyPi.
> 2) Allow the Inky 'saturation' to be configured globally, i.e. via the settings page.
> 3) Allow each plugin to set a 'saturation' value that makes sense for the image type they generate. This wouldn't affect non-Inky displays and the value would be ignored in those cases.
> 4) Implement a 'custom' dithering method and send a P type PNG to the Inky library. This could be global and/or per plugin type (i.e. dither based on the type of image).
> 
> I suggest doing (1), since this default seems to work better for 'photo' and 'web page' rendered images, and is quick.

thetaylormcrae:
This, by far, is the single greatest fix I've come across for my particular setup. It's bewildering a value of 0 would result in a better picture, but https://github.com/fatihak/InkyPi/issues/111#issuecomment-3215466577 makes a lot of sense!

@phattmatt : any idea if this setting would negatively impact other Pimoroni display types? Can't imagine it would, but always good to have that considered.

One other thing I came across was `Palette conversion failed` in the logs.

```shell
Jan 09 09:17:02 logi-display inkypi[1024]: /usr/local/inkypi/venv_inkypi/lib/python3.13/site-packages/inky/__init__.py:18: UserWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html. The pkg_resources package is slated for removal as early as 2025-11-30. Refrain from using this package or pin to Setuptools<81.
Jan 09 09:17:02 logi-display inkypi[1024]:   from pkg_resources import declare_namespace
[...]

Issue #111, opened 2025-05-07
Poor display (possible scaling problem) with Inky Impression 7.3 2025 Edition PIM773


menthol1979:
Hello all, 

Yesterday I received an Inky Impression 7.3 2025 Edition PIM773 display. It is somewhat different from the old I suppose, it has a whale on the back side and it's slightly larger than the one Fatihak uses in his videos : 

![Image](https://github.com/user-attachments/assets/f94aaff7-3236-4107-9c6d-9ee066df2f90)

I followed the instructions and it works absolutely fine, but the fonts, images etc are not as sharp as in demo videos.
More precisely, I get an image like this: 

![Image](https://github.com/user-attachments/assets/0218d4bb-0ee9-409e-af86-5c6446304324)

Sorry for the bad picture, lighting is poor now but I hope you get the big picture.

I know there was a modification to support the new displays but to my eyes this seems like a scaling problem. The installation is a clean one, following exactly fatihak's intructions from the site. 

Can anyone help me on this?
Thanks!
Nikos

Takaitra:
The PIM773 refreshes much faster than the older PIM667 but it dropped down to 6 colors from 7. The color it dropped is orange. I haven't looked at the code yet but I imagine updating the dithering logic or icon choices to avoid orange will fix the images.

I don't know about the text though.

menthol1979:
Apart from the color count reduction, all other specs (active pixels, resolution etc) seem identical. The dodgy black text, to my understanding, must be related to font or some kind of aliasing, don't know.
I am willing to try any attempts that could fix it though, if anyone has a better insight.

Takaitra:
I also have the PIM773 and see the same issue as you. I haven't worked on InkyPi code before but will dig in and see what I can learn. Give me some time though, I'm busy for the next week.

menthol1979:
> I also have the PIM773 and see the same issue as you. I haven't worked on InkyPi code before but will dig in and see what I can learn. Give me some time though, I'm busy for the next week.

Oh definitely, sure mate. No one's in hurry. I already enjoy it the way it is, just my comment to make it a tad better! Cheers!

Cogginshome:
I have the PIM774 13.3" it is missing the same colour. So if this gets fixed please add it aswell.

fatihak:
I'll work on setting up my 2025 Impression displays, will look into it this week

Takaitra:
[...]

PR #508, opened 2026-01-05
Modify Inky display saturation to 0 from the default of 0.5

phattmatt:
As per feature request https://github.com/fatihak/InkyPi/issues/502.

The PR modifies the Inky display driver's 'saturation' from the default of 0.5 to 0.

The 'saturation' parameter is not really saturation, but rather a parameter that affects how the colour palette is blended between the original values and 'desaturated' values.

Originally noticed and discussed here: https://github.com/fatihak/InkyPi/issues/111

I believe a 'saturation' value of 0 results in a better image displayed on the Inky Impression screens in the vast majority of situations.

### DIFF
--- a/src/display/inky_display.py
+++ b/src/display/inky_display.py
@@ -57,6 +57,7 @@ class InkyDisplay(AbstractDisplay):
         if not image:
             raise ValueError(f"No image provided.")
 
-        # Display the image on the Inky display
-        self.inky_display.set_image(image)
+        # Display the image on the Inky display - saturation set to 0 as it's not really saturation:
+        # https://github.com/pimoroni/inky/issues/225#issuecomment-3213935144
+        self.inky_display.set_image(image,saturation=0)
         self.inky_display.show()


roberthunt:
I think this may need to be configurable at least globally, if not per-per plugin. I did some testing (Inky Impression Spectra 13.3") with the image plugins with the `saturation` set to `0.5` and `0` but the images look significantly more 'washed out' and 'dull' with the setting at `0`. 

I know there isn't really a 'hardware config' section at the moment as it treats the display generically but it may be necessary to introduce this flexibility.

Attached an example photo of the display - not the best quality as there are some light reflections but enough to get an idea:

**saturation=0.5**
IMG_0350

**saturation=0**
IMG_0352

fatihak:
I agree with @roberthunt, I think this should be configurable in the settings page. Ideally want it to appear under the image settings if the display type is the spectra 6 display from Pimoroni. I can adding that sometime next week if you dont get to it.

fatihak:
Closing this in favor of https://github.com/fatihak/InkyPi/pull/530. I added an option for this under the Image Settings in the Settings page.

PR #510, opened 2026-01-07
Add note about color issues for Spectra 6 displays

kinto0:
Added a note regarding color issues with Spectra 6 displays.


fatihak:
Hi, the colors aren't necessarily 'wrong' and the linked issue doesn't provide any useful information. Something like this belongs more to the [troubleshooting guide](https://github.com/fatihak/InkyPi/blob/main/docs/troubleshooting.md) with details and steps.

Issue #167, opened 2025-06-21
Waveshare display not crisp


ian009:
Waveshare display not crisp

Description of display
7.3inch E Ink Spectra 6 (E6) Full Color E-Paper Display
E-Ink Display, Low Power Consumption, 800×480 Pixels, SPI Communication
7.3inch e-Paper HAT (E)
https://www.waveshare.com/product/displays/7.3inch-e-paper-hat-e.htm?sku=27875

installed latest version (as at 19 June 2025) using following command
sudo bash install/install.sh -W epd7in3e

Please see attached photo
Same problem observed on Raspberry Pi 4 & Raspberry Pi Zero 2W

![Image](https://github.com/user-attachments/assets/5c08c952-1852-48f6-9373-cf505c15c798)

cfhsu:
This is exactly what I found with my Waveshare 7.3" E6 display. I believe it comes from the diching process. I don't think we can really get rid of it unless the icon files have only the same 6 colors of a Spectra 6 display.

I have been tweaking with the settings a bit and found the following setting makes some improvement. Hope it helps.
Saturation: 1.2
Contrast: 1.0
Sharpness: 1.1
Brightness: 1.0

![Image](https://github.com/user-attachments/assets/3b92039a-c5c3-4b73-b549-3890a4e9fa37)

ian009:
Do you think that the Waveshare 7.3inch ACeP 7-Color E-Paper E-Ink Display Module, 800×480 Pixels will display correctly?

cfhsu:
The one color missing in E6 display is orange, which happens to be the dominant color of these weather icons. My guess is that it will be better, but not sure to what extent. The 7-color display has much longer refresh time and that's the reason I chose E6. Will have to dig more to see if I need to buy a 7-color one. Hope someone can share a picture of 7-color display here.

phattmatt:
InkyPi has no direct control over how the e-ink displays, or their drivers, work, I will close this issue.

Feel free to reopen, or raise a new issue, if you are still having problems.

kinto0:
two issues here:
- colors are wrong: in order to fix this, update your drivers with either of the changes for [epd7in3f ](https://github.com/waveshareteam/e-Paper/pull/411)or [epd7in3e](https://github.com/waveshareteam/e-Paper/pull/410) (no idea what the differences are, i'm on f)
[...]

PR #511, opened 2026-01-07
Attempt to fix waveshare orange color

kinto0:
hey, it's unfortunate that #167 exists after we paid ~100$ for a display. I'm hoping to change that. 

Can we turn oranges into the other colors this display supports?
- It seems like it could be done in `src/display/waveshare_display.py`

Or can we update the drivers themselves? (I don't know where the source code is for this)

In this PR, I update `src/display/waveshare_display.py`. 

Two parts left to this:
- How do we ensure this change is only applied to 6-color waveshare displays? I don't think this change is sufficiently gated but I'm not sure of the hardware options. Is it possible to fix?
- I would like assistance testing this. Currently I am making that change then reloading the display, but I don't notice a change. I've added log statements that should be hit and they are not. Reboot does not fix this.

### DIFF
--- a/src/display/waveshare_display.py
+++ b/src/display/waveshare_display.py
@@ -31,28 +31,24 @@ def split_image_for_bi_color_epd(image):
 
 def quantize_for_color_epd(image):
     """
-    Quantize image with extended palette for full-color e-paper displays.
+    Quantize image with correct palette for E6 (Spectra 6) full-color e-paper displays.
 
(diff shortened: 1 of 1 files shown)


kinto0:
fix is only possible in driver, not here:
- https://github.com/waveshareteam/e-Paper/pull/411
- https://github.com/waveshareteam/e-Paper/pull/410

PR #528, opened 2026-01-17
Adding driver for 4in e-paper

jimbogray:


jimbogray:
Adding driver for https://www.waveshare.com/4inch-e-paper-hat-plus-e.htm

fatihak:
Hi, the drivers are not committed to this repo but instead pulled during installation in the [install script](https://github.com/fatihak/InkyPi/blob/main/install/install.sh).

PR #531, opened 2026-01-18
Astronomy Picture of the Day, added options to fit image InkyPi Screen

philipp-leibner:
Hi, 

I added to minor options to the Astronomy Image of the Day Plugin

- Adjust to fit screen: Image is resized to fit the InkyPi screen using the device_config resolution info. By default a black background is used if borders are necessary
- Automatic background color: If turned on, the average color of the border of the image is calculated and used as background color 

I also adjust form-label CSS so that the toggles align 

grafik

Result for today (18.1.26)
grafik


fatihak:
Hi, thanks for the PR! I would prefer the options be renamed/defined similar to other image related plugins to be consistent:
Screenshot 2026-01-24 at 7 40 17 PM

philipp-leibner:
Makes sense! I'll adjust it :)

philipp-leibner:
I made the change. I also added another option to add the image title and copyright to the image in the lower right corner including fontsize and padding settings.

grafik

grafik

phattmatt:
Hi @philipp-leibner,

Thanks for your contribution!

I'm reviewing your PR and the improvements look good.

Just a few notes:

1) I don't think you need to include the "src/config/device_dev.json" file. You may have added it inadvertently.
2) Can you confirm the "src/static/styles/main.css" changes are necessary? This change would affect all plugins and if not needed would be easier to test without it.
3) I don't think your [font loading code](https://github.com/fatihak/InkyPi/pull/531/changes#diff-c6dec1f807ffb88b70064c12fa553e5b178746c602418fc17e3662779346a406R96) is working as on my InkyPi instance it's generating the following error:

```
Aug 19 13:31:06 inkypi-dev inkypi[42815]: 13:31:06 - ERROR - plugins.apod.apod - Failed to load font, using default font.
```

Matt.

philipp-leibner:
Hi Matt,

without the change in `main.css` the labels would look like this:
grafik
I wanted to improve the alignment (as above) of the labels. Yes it affects all plugins, I personally think it looks cleaner, but if you prefer the original version I can remove the changes.

Adding "src/config/device_dev.json" was a mistake, did not see that.

Regarding the font issue, I fixed that issue. Now the font path is build properly from the project root.

PR #512, opened 2026-01-07
Add hardware button support and API keys management UI

v-murygin:
This PR adds comprehensive hardware button support for Inky Impression displays and introduces a web-based API keys management interface.

### 🔘 Hardware Button Support
- Physical GPIO button handling for Raspberry Pi (buttons A, B, C, D)
- Virtual button controls in web UI with visual feedback
- Support for both short press and long press actions
- Configurable button actions via web UI
- Plugin-specific button handlers (e.g., Colors plugin responds to B/C for color cycling)

**Default button mapping:**
| Button | Short Press | Double Press | Long Press |
|--------|-------------|--------------|------------|
| A | Refresh display | - | - |
| B | - | - | Previous plugin |
| C | - | - | Next plugin |
| D | - | - | Shutdown |

Button actions are fully configurable via Settings → Button Actions section.
### Display Refresh Indicator
- Loading overlay on preview image during e-ink refresh

###  API Keys Management
- New `/api-keys` page for managing environment variables
- Secure value masking with reveal toggle
- Direct `.env` file editing without SSH access
- Hot reload of environment variables

###  Settings Improvements
- Redesigned Settings page with collapsible sections
- Hardware Settings: GPIO pin configuration, button visibility toggle
- Button Actions: configure short/double/long press actions for each button

### ℹ️ System Info Popover
- Quick access to device information from Settings header

### 🔲 Plugin Grid Improvements  
- Plugin names now displayed under icons
- Grid/List view toggle with preference saved to localStorage
- Drag-and-drop plugin reordering
- Custom order persisted to device config

### New Module: `src/buttons/`
- `abstract_button_handler.py` - Base class and enums (ButtonID, PressType)
- `button_manager.py` - Event routing and action execution
- `inky_button_handler.py` - GPIO implementation using gpiozero
- `mock_button_handler.py` - Development/testing stub

### Plugin Button API
Plugins can now implement `on_button_press(button_id, press_type, device_config)` to handle button events.
```python
def on_button_press(self, button_id, press_type, device_config) -> bool:
    if button_id == ButtonID.C and press_type == PressType.SHORT:
        self.current_index = (self.current_index + 1) % len(self.items)
        return True  # Handled, triggers display refresh
    return False  # Not handled, falls through to configured action
```

## Dependencies

Added to `requirements.txt`:
- `gpiozero>=2.0.1`
- `lgpio>=0.2.2.0`

## Screenshots
Screenshot 2026-01-07 at 9 43 05 PM
[...]

PR #532, opened 2026-01-19
Add hardware button support (part 1)

v-murygin:
> **Note:** This PR is part 1 of the split from #512 (as requested). Contains hardware button support only. Part 2 (UI improvements & API keys) will follow separately.

This PR adds hardware button support for e-ink displays with GPIO controls.

### 🔘 Hardware Button Support
- Physical GPIO button handling for Raspberry Pi (buttons A, B, C, D)
- Virtual button controls in web UI with visual feedback
- Support for short press, double press, and long press actions
- Configurable button actions via Settings → Button Actions
- Enable/disable toggle in Settings → Hardware Settings
- Configurable GPIO pins (default: A=5, B=6, C=16, D=24 for Inky Impression)

**Default button mapping:**
| Button | Short Press | Long Press |
|--------|-------------|------------|
| A | Refresh display | - |
| B | Plugin action | Previous plugin |
| C | Plugin action | Next plugin |
| D | - | Shutdown |

###  Settings Improvements
- Collapsible sections for better organization
- Hardware Settings: GPIO pin configuration, button visibility toggle
- Button Actions: configure actions for each button/press type

###  Plugin Button API
Plugins can implement `on_button_press()` to handle button events:
```python
def on_button_press(self, button_id, press_type, device_config) -> bool:
    if button_id == ButtonID.C and press_type == PressType.SHORT:
        self.next_item()
        return True  # Handled, triggers refresh
    return False  # Falls through to configured action
```

###  Photo Frame Plugin
Test plugin demonstrating button navigation (B/C for prev/next photo).

### New Module: `src/buttons/`
- `abstract_button_handler.py` - Base class and enums
- `button_manager.py` - Event routing and action execution  
- `gpio_button_handler.py` - GPIO implementation using gpiozero
- `mock_button_handler.py` - Development/testing stub

## Dependencies
- `gpiozero>=2.0.1`
- `lgpio>=0.2.2.0`

<details>
  <summary>
    <h3 style="display:inline">📸 UI Screenshots </h3>
  </summary>

  <br>
  Web UI Settings
  <br>
  Mobile View Settings
  
</details>


src/buttons/button_manager.py
```diff
@@ -0,0 +1,334 @@
...
+        except ValueError:
+            next_name = playlist_names[0]
+        
+        playlist_manager.active_playlist = next_name
+        self._device_config.write_config()
+        
+        logger.info(f"Switched to playlist: {next_name}")
+        self._refresh_task.signal_config_change()
+    
+    def _toggle_cycle(self):
```
fatihak: Not sure what this button is for, _auto_cycle_enabled doesnt seem to be used anywhere

src/buttons/button_manager.py
```diff
@@ -0,0 +1,334 @@
...
+        """Toggle automatic plugin cycling."""
+        self._auto_cycle_enabled = not self._auto_cycle_enabled
+        status = "enabled" if self._auto_cycle_enabled else "disabled"
+        logger.info(f"Auto-cycle {status}")
+    
+    def _show_info(self):
```
fatihak: This action doesn't update the screen for me, likely bc it instantiates a new DisplayManager instance. I dont think we should be instantiating a new display manager here but instead pass it into the button manager in inkypi.py

src/buttons/button_manager.py
```diff
@@ -0,0 +1,334 @@
...
+            dimensions = self._device_config.get_resolution()
+            if self._device_config.get_config("orientation") == "vertical":
+                dimensions = dimensions[::-1]
+            
+            width, height = dimensions
+            img = Image.new("RGB", dimensions, (30, 30, 30))
+            draw = ImageDraw.Draw(img)
+            
+            try:
+                font = get_font("DejaVuSans", 24)
```
fatihak: I'm seeing this warning: 
Requested font not found: font_name=DejaVuSans

We should just use one of the existing fonts committed to the repo

src/static/styles/main.css
```diff
@@ -122,16 +124,18 @@ body {
 /* Current image styling */
 .image-container {
     position: relative;
-    width: 300px;
+    width: fit-content;
     margin: 0 auto;
+    padding: 6px;
     cursor: pointer;
     transition: opacity 0.3s ease;
+    background-color: #1a1a1a;
```
fatihak: Remove, this adds a black border to the image which doesnt look nice in light mode and isnt noticeable in dark mode

fatihak (changes requested):
Thanks for the PR! This is a really nice implementation overall and very close to how I was thinking about adding button support, especially with the settings flow and GPIO handling.

A few items of feedback:
- Plugin button API: I don’t think this is really necessary. For most cases (for example, “next” image in an album or library), a refresh already achieves the same result. Exposing button handling inside plugins will likely lead to requests for plugin-specific button settings which adds complexity when defining a separate plugin instance with different settings is the expected use case.
- Button actions: I think the most useful/requested behavior will be mapping buttons to show a specific plugin. For example, button A shows a calendar, button B shows a clock, etc. This doesn’t need to be added as part of this PR since it should be straightforward to layer on later with the existing structure.
- Added plugins: I’m guessing the added plugins were to demonstrate the button actions/testing but they can be removed. You can add them as third plugin plugins which were recently added: https://github.com/fatihak/InkyPi/wiki
- Concurrent button actions: Right now, pressing a button multiple times with some delay can trigger multiple actions before the first one finishes. I think we should prevent new button actions from executing while another is in progress. Otherwise this can lead to race conditions especially with resource-heavy plugins like refreshing weather or calendar plugins at the same time.

v-murygin:
@fatihak  Thanks for the detailed feedback!  I've addressed almost all points in the latest commits.                                                                                                                                                                                                                                            

Re: Plugin button API - I respectfully disagree here.  The API solves a problem that refresh alone can't handle.
Example (Photo Frame plugin): User has 10 photos in one instance, buttons B/C navigate between them. Simple refresh won't work - it would show the same photo since there's no state change. Creating 10 separate plugin instances for 10 photos would be poor UX. 
The API is simple (plugin returns True if handled, False to fall through), opt-in, and enables interactive plugins like galleries and slideshows without forcing multiple instances.   

Note: I've also added concurrent action protection at both client and server levels to prevent race conditions when buttons are pressed rapidly. 
PS: I tweaked the button UI - users can now choose where to display buttons on the home page: bottom, top, left, or right side.

RobinWts:
@v-murygin is absolutely right here. If you expect people to contribute by building plugins, you need APIs, regs, hooks, and callbacks for them to interact with as much as possible. Of course, it adds complexity, but it does so in a controlled way.
Expose more functionality to the plugins, and you will probably be surprised how people use that. If you don’t, this project will be dependent on core integration, which is inherently slow. If you want plugins, plugins, plugins (as stated somewhere), you need to grant them the possibilities to do stuff…

v-murygin:
@fatihak  Any updates? I would like to finish this PR :smiling

schmitzkr:
I tried this and it seems to work well (pi zero with inky impressions display), thanks for sharing your work!

PR #537, opened 2026-01-20
Add avif to supported images

boidolr:
Support was added with pillow v11.3.0


fatihak (approved):

PR #544, opened 2026-01-24
Add plugin grid improvements and API keys management interface (part 2)

v-murygin:
> **Note:** This PR is part 2 of the split from #512 (as requested). Contains UI improvements and API keys management. Part 1 (hardware buttons) is #532.

This PR adds plugin grid improvements and API keys management interface.

### 🔲 Plugin Grid Improvements
- Plugin names now displayed under icons
- Grid/List view toggle with preference saved to localStorage
- Drag-and-drop plugin reordering
- Custom order persisted to device config

### 🔑 API Keys Management
- New `/api-keys` page for managing environment variables
- Secure value masking with reveal toggle
- Direct `.env` file editing without SSH access
- Hot reload of environment variables
- Preset buttons for common API keys (OpenWeatherMap, OpenAI, NASA, Unsplash, GitHub)

<details>
  <summary>
    <h3 style="display:inline">📸 UI Screenshots</h3>
  </summary>
  <br>
  
Screenshot 2026-01-23 at 9 48 54 PM
Screenshot 2026-01-23 at 9 49 12 PM
Screenshot 2026-01-23 at 9 50 08 PM

  
</details>

### DIFF
--- a/src/blueprints/apikeys.py
+++ b/src/blueprints/apikeys.py
@@ -1,4 +1,5 @@
 from flask import Blueprint, request, jsonify, current_app, render_template
+from dotenv import dotenv_values
 import os
 import re
 import logging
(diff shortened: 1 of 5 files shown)


fatihak:
Thanks for the PR! Overall looks good and I like the changes here. A few items of feedback:
- I dont think we should provide the actual values of the api keys or have the ability to view them in the UI. Since there is no authentication it can be viewed/accessed by anyone on the same network. 
- I think the API Keys button on the home page can be moved up to the top level buttons, next to the playlist button

v-murygin:
@fatihak 
Thanks for the review! I addressed all the feedback — removed the reveal button for security, moved the 🔑 to header, fixed the button text and moved styles to main.css.

Also thinking we could move theme toggle and API Keys button into Settings, or even make it a collapsible section there instead of a separate page — to keep the header cleaner. But probably better to do after #532 merges to avoid conflicts. What do you think?

PR #559, opened 2026-01-29
Calendar: Add login info, attendance color and time intervals for now indicator

terentyevalexey:
Added login information for simple auth requiring calendars, attendance color (for attendees or organizer) and new time intervals for Now Indicator (5, 10, 15, 30, 60 minutes)

Possiblee problems: One login/pw for all calendars, possibly unsafe pw handling, same color for attendee/organizer; not all time intervals covered.


fatihak:
Thanks for the PR, these are interesting features. However, I don’t feel they add functionality that needs to be committed to the project.

PR #566, opened 2026-02-02
Add API editing/entering into GUI.

snowsecure:
Used CODEX to add a way to interact with and edit API keys needed while inside the GUI. I hated having to use a .env file.

### DIFF
--- a/src/button_handler.py
+++ b/src/button_handler.py
@@ -1,0 +1,102 @@
+import logging
+import threading
+import pytz
+from datetime import datetime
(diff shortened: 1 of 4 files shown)


fatihak:
Hi, and thanks for the PR! If you pull the latest changes an API Keys Management page has already been added as part of [#544](https://github.com/fatihak/InkyPi/pull/544). The existing solution is more modular and allows any key/values to be added rather than hardcoding existing key names. Feel free to pull the latest changes and give it a try.

PR #568, opened 2026-02-02
Handle background image color for grayscale images

boidolr:
This PR fixes loading of grayscale images with padding. Previous error:
> inkypi[1148]:   File "/home/inkypi/src/plugins/image_folder/image_folder.py", line 75, in generate_image
> inkypi[1148]:     img = ImageOps.pad(...)
> inkypi[1148]:   File "/home/inkypi/.venv/lib/python3.14/site-packages/PIL/ImageOps.py", line 363, in pad
> inkypi[1148]:     out = Image.new(image.mode, size, color)
> inkypi[1148]:   File "/home/inkypi/.venv/lib/python3.14/site-packages/PIL/Image.py", line 3118, in new
> inkypi[1148]:     return im._new(core.fill(mode, size, color))
> inkypi[1148]:                    ~~~~~~~~~^^^^^^^^^^^^^^^^^^^
> inkypi[1148]: TypeError: color must be int or single-element tuple

When loading grayscale images (mode: `L`), colors from `RGB` mode as background/padding will lead to errors. Grayscale in general is supported (see https://github.com/fatihak/InkyPi/blob/main/src/utils/image_utils.py#L67).

Resolving `ImageColor` requires `str` arguments instead of tuples (which may be the result of the function), see https://pillow.readthedocs.io/en/stable/reference/ImageColor.html#PIL.ImageColor.getcolor

### DIFF
--- a/src/plugins/ai_image/ai_image.py
+++ b/src/plugins/ai_image/ai_image.py
@@ -3,7 +3,7 @@ from openai import OpenAI
 from PIL import Image
 from io import BytesIO
 import base64
-import requests
+from utils.http_client import get_http_session
 import logging
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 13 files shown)


fatihak:
Looks good, thanks for the fix!

PR #572, opened 2026-02-04
Add a check for Google Chrome's default macOS installation path + test

valentincosson:
### Description
This PR improves the browser discovery logic by adding the default macOS installation path for Google Chrome /Applications/Google Chrome.app/Contents/MacOS/Google Chrome

### Changes
- **src/utils/image_utils.py**: Updated [_find_chromium_binary](cci:1://file:///Users/valentin/Documents/code.nosync/perso/InkyPi/src/utils/image_utils.py:108:0-123:15) to check the standard macOS Chrome location as a fallback.
- **tests/utils/test_image_utils.py**: Added a minimalist unit test to verify that the macOS path is correctly discovered when other options are unavailable.

### Verification Results
- [x] Verified binary discovery on macOS (Chrome found at expected path).
- [x] Ran the full test suite (including the new test).

### Impact
This change improves the experience for contributors developing on macOS without requiring manual PATH configuration for Chrome.

### DIFF
--- a/src/utils/image_utils.py
+++ b/src/utils/image_utils.py
@@ -114,6 +114,13 @@ def _find_chromium_binary():
         if path:
             logger.debug(f"Found browser binary: {candidate} at {path}")
             return candidate
+    
+    # Check default macOS location for Google Chrome
+    macos_chrome = "/Applications/Google Chrome.app/Contents/MacOS/Google Chrome"
+    if os.path.exists(macos_chrome):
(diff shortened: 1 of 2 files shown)


fatihak:
Thanks for the PR. I’m a bit hesitant to hardcode an OS-specific browser path here.
I’m wondering if a configurable approach like an environment variable might be a better approach, so this can be reused across different OSes and chrome locations.

PR #576, opened 2026-02-04
Plugin Management UI

RobinWts:
# Plugin Management UI

Since there was no "easy" way to install and uninstall 3rd party plugins I created a basic Web UI for it.
Changes to core file are minimal, except for `install/cli/inkypi-plugin` this is used for install/uninstall via webUI to keep it single source of thruth.
Layout matches the current core functions.
Tested in -dev and on device. 
Includes the MacOS Chromium path from Mac-Desktop Chrome.

Sorry for this second PR, the first included some rubbish.

Bildschirmfoto 2026-02-04 um 12 49 49
Bildschirmfoto 2026-02-04 um 12 50 04

## Features

- **Manage plugins** button on main screen (between dark mode and API Keys)
- List installed third-party plugins with delete buttons
- Install from GitHub.com repository URL (validates plugin structure: folder with `render/` and `plugin-info.json`)
- Success message with "Reload page" button after install/uninstall

## Files Changed

- `install/cli/inkypi-plugin` - Added `install-from-url <repo_url>` command corrected info output and tweaked the logic to work with UI feedback AND in -dev, added record of GIT URL if not in config to id 3rd part plugins.
- `src/blueprints/plugin_manage.py` - New blueprint: GET page, POST install/uninstall
- `src/inkypi.py` - Register plugin_manage blueprint
- `src/templates/inky.html` - Add "Manage plugins" button
- `src/templates/plugin_manage.html` - New template: plugin list, install form

### DIFF
--- a/src/blueprints/plugin_manage.py
+++ b/src/blueprints/plugin_manage.py
@@ -159,3 +159,60 @@ def uninstall_plugin():
                       result.returncode, output.strip())
     
     return jsonify({"success": True})
+
+
+@plugin_manage_bp.route("/manage-plugins/update", methods=["POST"])
+def update_plugin():
+    """Update a third-party plugin by reinstalling from its repository. Only allows plugins with repository set."""
+    data = request.get_json() or {}
+    plugin_id = (data.get("plugin_id") or "").strip()
+
+    if not plugin_id:
+        return jsonify({"success": False, "error": "plugin_id is required"}), 400
+
+    third_party = _third_party_plugins()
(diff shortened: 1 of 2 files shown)


fatihak:
Thanks for taking the time to put this together. After reviewing it, I don’t think this is something I want to merge into the project right now, so I’m going to close the PR.

I appreciate the effort and interest in contributing.

RobinWts:
I see why you wouldn't. So i made a 3rd-party plugin out of it.

PR #584, opened 2026-02-09
Dynamic font loading

RobinWts:
This introduces dynamic font loading to the InkyPi core. For full description see: [/docs/fonts.md](https://github.com/RobinWts/InkyPi/blob/clean_DynamicFontLoading/docs/fonts.md).

It maintains full backward compatibility still using the hardcoded fonts and they keep prioity even when dynamic loading finds them somwhere else with the same name.

It makes all fonts available for everyone and includes fonts that come with installed plugins.

It makes it easy for plugins to bring their own fonts without needing to modify core files.
(without the need to modify plugin mechanism, it just scans all plugins for font dirs and includes them)

It scans for fonts the first time they are requested and keeps them available for rendering (base plugin) and settings.

The internal API remains unchanged, so nothing else needs to be updated.

It uses fonttools to extract meta data. 

It comes with extensive fall back and overwrite logic to be robust and flexible.

It's easy to use: just drop fonts in the fonts dir or your plugin's fonts dir, restart the service and be done - no further code changes required.

All in all it makes developing plugins easier and InkyPi more 3rd-party-plugin-friendly.

### DIFF
--- a/docs/building_plugins.md
+++ b/docs/building_plugins.md
@@ -98,8 +98,15 @@ plugins/{plugin_id}/
     ├── icon.png                # Plugin icon
     ├── settings.html           # Optional: Plugin settings page (if applicable)
     ├── render/                 # Optional: If generating images from html and css files, store them here
+    ├── fonts/                  # Optional: Plugin-specific fonts (automatically discovered)
+    │   ├── CustomFont.ttf
+    │   └── CustomFont-Bold.ttf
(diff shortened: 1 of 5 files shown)


RobinWts:
For info: I force pushed this, to sync it with the requirements bump from 13th Feb 26 since I changed those files as well and removed one merge commit I accindently included.

PR #592, opened 2026-02-12
Bump required versions

boidolr:
Bring these in line with current versions, including a number of security updates.


fatihak (approved):
Thanks for the PR

PR #601, opened 2026-02-15
feat(plugins): add finance tracker and daily quote plugins

BloodAkatsuki:
## Summary
- **Finance Tracker**: Crypto, stocks & commodities with sparkline charts. Uses CoinGecko API (crypto) and Yahoo Finance (stocks/gold). Configurable currency, timeframe, and asset list.
- **Daily Quote**: Displays random quotes from ZenQuotes API with built-in fallback quotes. Configurable font size and responsive typography.

## Details
- No new dependencies required (uses only `requests`)
- Both plugins follow existing plugin patterns (BasePlugin, render_image, style_settings)
- B&W optimized for e-ink displays
- Supports both landscape and portrait orientations

## Test plan
- [x] Finance Tracker: Add crypto (bitcoin, ethereum) and verify prices + sparklines render
- [x] Finance Tracker: Add stock and commodity (gold, silver) symbols
- [x] Daily Quote: Verify quote loads and displays centered
- [x] Test with vertical orientation


IrishJohnnyP:
Thank you for putting this together!   A couple of issues.  When querying bitcoin, Dow jones index and the Nasdaq index there is an error.  

Also putting this as a 3rd party plugin is easy and works for both of these.

fatihak:
Thanks for the PR!
As part of a recent change, we’re no longer accepting new plugins directly in the main repository. Instead, InkyPi now supports third-party plugins.
You can check out the updated docs here: [Publishing a Third Party Plugin]( https://github.com/fatihak/InkyPi/blob/main/docs/building_plugins.md#publishing-a-third-party-plugin). In short, you can move this plugin into its own repository and then add it to the plugin list on the [the wiki](https://github.com/fatihak/InkyPi/wiki). 
Let me know if you run into any issues.

PR #625, opened 2026-03-16
fix(immich): prevent OOM crash on large images and fix video asset handling

iAndrea:
## Problem

The Immich plugin was downloading full-resolution originals (up to 24MB+) on
every refresh, causing OOM kills on low-RAM devices (e.g. Pi Zero). Additionally,
video assets in albums could be selected and passed to `Image.open()`, causing
a crash, and a missing `import requests` caused a `NameError` that swallowed
the real error in the image loader.

### `image_album.py`
- **Download preview instead of original** — replace the hard-coded
  `/api/assets/{id}/original` URL with a version-aware resolver that probes
  the server once (streaming GET, no body downloaded) and caches the working
  endpoint for the lifetime of the provider:
  1. `/api/assets/{id}/thumbnail?size=preview` — current Immich API
  2. `/api/assets/{id}/thumbnail` — older versions without `size` param
  3. Falls back to `/original` with a warning if both fail
- **Switch asset fetch to album endpoint** — replace the paginated
  `POST /api/search/metadata` loop with a single `GET /api/albums/{id}`,
  with client-side `type == "IMAGE"` filtering to exclude video assets
- **README** — add `asset.view` to the required API key permissions

### `image_loader.py`
- **Fix `NameError` in exception handlers** — `requests` was referenced in
  `except requests.exceptions.RequestException` but never imported
- **Eliminate redundant memory copy in `_load_from_url_fast`** — replace
  `BytesIO(response.content)` (two copies of compressed bytes in memory) with
  `response.raw` piped directly into PIL; `decode_content=True` handles
  gzip/deflate transparently

## Required API key permissions

`asset.view` **(NEW)**, `asset.read`, `asset.download`, `album.read`

### DIFF
--- a/src/plugins/image_album/README.md
+++ b/src/plugins/image_album/README.md
@@ -3,6 +3,6 @@
 <a href="https://immich.app/">Link to Immich</a>
 
 Create an album in Immich<br>
-Create an API Key with the following permissions: asset.read, asset.download, album.read<br>
+Create an API Key with the following permissions: asset.view, asset.read, asset.download, album.read<br>
 Store the key in the .env file with IMMICH_KEY=1234<br>
 
--- a/src/plugins/image_album/image_album.py
+++ b/src/plugins/image_album/image_album.py
@@ -10,12 +10,61 @@ logger = logging.getLogger(__name__)
 
 
 class ImmichProvider:
+    # Ordered from newest to oldest.  The first URL that answers 200 to a HEAD
+    # request is stored in _preview_url_template and reused for all subsequent
+    # asset fetches so we only pay the probe cost once per provider lifetime.
+    #
+    # Current Immich API (viewAsset operationId):
+    #   GET /api/assets/{id}/thumbnail?size=preview
+    # Older versions (no size param, returns preview JPEG by default):
+    #   GET /api/assets/{id}/thumbnail
+    _PREVIEW_URL_TEMPLATES = [
+        "{base}/api/assets/{id}/thumbnail?size=preview",  # Immich current API
+        "{base}/api/assets/{id}/thumbnail",               # Immich older (no size param)
+    ]
+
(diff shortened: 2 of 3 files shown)


wildmor:
I can confirm this PR fixes the OOM issue on Raspberry Pi Zero 2 W.
My setup: RPi Zero 2 W (512MB RAM) + InkyPi + Image Album plugin pulling photos from Immich. The album contains a mix of Android and iPhone (HEIC) photos.
Before this fix, InkyPi was regularly killed by earlyoom — Python was spiking to ~126MB RAM during image processing, which on a 512MB device left no headroom for anything else. HEIC photos from iPhone were particularly painful since they were served as full-resolution originals.
After applying this PR, no OOM kills since. Great fix, hope it gets merged soon!

RoliMG:
Hi, original author of the plugin here. Tested your changes and would approve if I could.

iAndrea:
@RoliMG I think that's a shared feeling about this repo... unfortunately. 😞

phattmatt:
Hi @iAndrea @RoliMG @wildmor 

I am unable to test these changes myself, as I don't have access to an Immich instance.

However, I am willing to merge this PR if several people can confirm this PR is working, and has no breaking changes.

The more feedback we get, the more likely it is I will merge.

Thanks.

Matt.

RoliMG:
Hi, I have it running now for ~4 weeks and didn't have any issues. Everything works well so far.

phattmatt (approved):
Hi @iAndrea,

I managed to find a Immich demo instance for me to run some basic tests against:

https://demo.immich.app/

Looks like your changes are working well, and I can't see any regressions.

The positive feedback from @RoliMG and @wildmor gives me enough confidence to merge your PR.

Thanks all for your contributions including the original plugin, this PR, and the feedback provided.

Matt.

PR #723, opened 2026-06-28
AI-Image plugin.  Add support for GPT-Image-2 model.  Remove support for deprecated Dalle models.

rgelb:
Changes for the AI-Image plugin.

Removed support for Dall-e-2 and Dall-e-3 models.  
Added support for GPT-Image-2 model (current OpenAI image model)

### DIFF
--- a/src/plugins/ai_image/ai_image.py
+++ b/src/plugins/ai_image/ai_image.py
@@ -3,14 +3,13 @@ from openai import OpenAI
 from PIL import Image
 from io import BytesIO
 import base64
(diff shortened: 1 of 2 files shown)


phattmatt (approved):
Hi @rgelb,

I've tested your changes and they look good.

The AI Image plugin can now be used with the current models.

The only downside is any existing plugin instances using the old models will need to be edited and a new model selected, but since they will be failing anyway I don't see this as a problem.

Thanks for you contribution!

Matt.

PR #735, opened 2026-08-05
Add iCloud Shared Albums plugin

flavorplus:
## Summary

Adds an iCloud Photos plugin that displays images from a public iCloud Shared Album without requiring Apple authentication.

- Selects photos randomly without repeating them until the album has been exhausted
- Reconciles added, removed, and replaced photos between refreshes
- Supports crop-to-fill and scale-to-fit with blurred or solid-color backgrounds

## Implementation

The plugin is under `src/plugins/icloud_photos`.

Viewed-photo state is stored in the plugin instance settings and is updated only after an image has downloaded and rendered successfully. This is to keep track of already displayed photos.

All external requests have explicit timeouts. Shared-album and signed asset URLs are not logged by the plugin.

## Testing

- Full test suite: 67 passed
- Added 43 mocked and parameterized iCloud test cases
- Opt-in live integration test passes against a public iCloud Shared Album
- Live test verifies the current `webstream` and `webasseturls` schemas and downloads an image through `AdaptiveImageLoader`
- Tested with a physical Pimoroni Inky 7.3inch display and Raspberry Pi Zero 2 W running on Debian 13.6 Trixie.

The live integration test is skipped by default to keep normal test runs deterministic.

## Notes

Apple's public Shared Album endpoints are undocumented and could change. The opt-in live test is included to detect schema or download-flow changes without making upstream CI dependent on an external service.


phattmatt:
Hi @flavorplus, 

Thanks for the contribution, the plugin looks great.

InkyPi has introduced support for 3rd party plugin installs and as such will probably not accept new plugins into the core project:

https://github.com/fatihak/InkyPi/blob/main/docs/building_plugins.md#publishing-a-third-party-plugin

https://github.com/fatihak/InkyPi/wiki/3rd-Party-Plugins

I hope you consider re-working your plugin and publishing it that way.

In the meantime I will close this PR, feel free to re-open, or open a new one if you feel this has been closed in error.

PR #692, opened 2026-05-20
Add Unsplash illustration collection mode (and fix import)

k4m454k:
This PR updates the Unsplash plugin so it can load not only regular photos, but also Unsplash `illustration` assets from collections.

Previously, the plugin used `/photos/random` when no search query was provided. That works for photo collections, but Unsplash does not return `asset_type: illustration` items through that endpoint. Collections that contain only illustrations therefore failed with `No photos found`, even though `/collections/{id}/photos` returned valid items.

The plugin now supports an explicit content type setting:

- `Photos` keeps the existing behavior.
- `Illustrations from collections` loads only `asset_type: illustration` items.
- `Photos and illustrations from collections` loads both photos and illustrations.

**Key Changes**
- Added `asset_type` handling in `src/plugins/unsplash/unsplash.py`.
- Preserved existing photo behavior via:
  - `/search/photos` when `search_query` is set.
  - `/photos/random` otherwise.
- Added collection-based loading for `illustration` and `mixed` modes using:
  - `/collections/{collection_id}/photos`
- Added support for multiple comma-separated collection IDs.
- Added filtering by `asset_type`.
- Added image URL fallback order:
  - selected size (`regular` on low-resource devices, `full` otherwise)
  - `regular`
  - `small`
  - `full`
  - `raw`
- Added clear runtime errors for invalid combinations:
  - `illustration` or `mixed` without collections.
  - `illustration` or `mixed` with `search_query`.
  - collections with no matching assets.
- Updated `src/plugins/unsplash/settings.html` with a new `Content Type` select.
- Added `tests/test_unsplash_plugin.py` covering the new behavior.

**Behavior Notes**
- Illustration support intentionally uses the documented collection photos endpoint rather than an undocumented `/illustrations/random` endpoint.
- Search remains photo-only because Unsplash’s documented public search endpoint is `/search/photos`.
- In collection-based `illustration` and `mixed` modes, `orientation` is passed through, but `color` is not sent because the collection photos endpoint does not document a `color` parameter.
- No separate SVG/vector conversion is needed: Unsplash returns raster image URLs in `urls.regular`, `urls.full`, etc., and the existing image loader already downloads and converts them through Pillow.

**Testing**
Ran:

```bash
PYTHONDONTWRITEBYTECODE=1 python3 -m unittest tests.test_unsplash_plugin -v
```

Result:

```text
Ran 5 tests in 0.054s

OK
```

Covered scenarios:
- Existing photo random endpoint behavior is preserved.
[...]

PR #737, opened 2026-08-08
Restore Unsplash API request timeout

TonyTTTT:
## Summary

Restore the 30-second timeout for Unsplash API requests while continuing to use InkyPi's shared HTTP session.

This also restores the `requests` import required by the existing `requests.exceptions.RequestException` handler.

## Problem

PR #545 added explicit 30-second request timeouts to prevent network requests from hanging indefinitely.

PR #427 later migrated the Unsplash plugin to the shared HTTP session, changing:

```python
requests.get(url, params=params, timeout=30)
```

to:

```python
session.get(url, params=params)
```

That unintentionally removed the timeout. It also removed the `requests` import even though the existing exception handler still references `requests.exceptions.RequestException`.

Without an explicit timeout, an Unsplash refresh can remain blocked when connectivity, DNS, or routing is interrupted. Because playlist refreshes run on a single background thread, this can prevent subsequent scheduled updates from running.

## Changes

* Add `timeout=30` to the Unsplash API request.
* Restore `import requests` for the existing exception handler.
* Retain the shared HTTP session and connection pooling introduced by PR #427.

When the request times out, Requests raises a `Timeout` exception, which inherits from `RequestException`. The existing handler therefore logs the failure and raises `RuntimeError`, allowing the refresh task to handle the failed attempt instead of remaining blocked.

## Testing

* `python3 -m py_compile src/plugins/unsplash/unsplash.py`
* `git diff --check`

Both checks completed successfully against the current `main` branch with this patch applied.

## Related pull requests

* #545 — originally added request timeouts to prevent hangs.
* #427 — unintentionally removed the Unsplash timeout during the shared-session refactor.
* #692 — restores the missing `requests` import, but its Unsplash API requests still lack explicit timeouts.

### DIFF
--- a/src/plugins/unsplash/unsplash.py
+++ b/src/plugins/unsplash/unsplash.py
@@ -3,6 +3,7 @@ from utils.image_loader import _is_low_resource_device
 from utils.http_client import get_http_session
 import logging
 import random
+import requests
 
 logger = logging.getLogger(__name__)
 
@@ -60,7 +61,7 @@ class Unsplash(BasePlugin):
         try:
             logger.debug("Fetching image from Unsplash API...")
             session = get_http_session()
-            response = session.get(url, params=params)
+            response = session.get(url, params=params, timeout=30)
             response.raise_for_status()
             data = response.json()
 


phattmatt (approved):
Hi @TonyTTTT,

Thanks for your fix!

Looks good to me, so I'll merge.

Matt.

PR #738, opened 2026-08-09
Fix playlist plugin refresh intervals

Rs4anti:
## Fix playlist plugin refresh intervals

This change fixes refresh scheduling for playlist plugin instances with their own refresh interval.

### Changes

* The refresh task now wakes up according to the currently displayed plugin's refresh interval.
* Plugin-specific refresh intervals can trigger a refresh without waiting for the global playlist cycle interval.
* This fixes plugins such as the clock, which need to regenerate their image periodically while remaining the currently displayed playlist item.

### Example

With:

* `plugin_cycle_interval_seconds = 600`
* Clock plugin `interval = 60`

the clock is refreshed every 60 seconds while it remains the active plugin, while playlist rotation continues to use the global cycle interval.


saulob:
I already have a PR open for this issue since March 19, #629 

It addresses the scheduled refresh behavior, including plugin-specific refresh intervals

Maybe we can compare both approaches and consolidate the fixes into one PR

Rs4anti:
Yep, fine for me.

I have tested personally only witch only, have you tested it also for other
plugins? Because I will need it for calendar plugin also.

Greetings

saulob:
Yes, my PR works with all plugins, not just the clock plugin

If you can test my PR with the Calendar plugin as well, I would really appreciate it :)

Thanks

Rs4anti:
Yes I will ,

I'll notify you when done.

Thanks!

PR #103, opened 2025-04-25
Added Waveshare ePaper Device Support

acscpt:
Have added support for [Waveshare e-paper devices](https://github.com/waveshareteam/e-Paper/blob/master/RaspberryPi_JetsonNano/python/lib/waveshare_epd/).

1. Added an optional parameter for install `-W <device>` where <device> is the device name as per the driver name, e.g. epd7in3f for the 7-color ACEP screen.  This also corresponds to the actual python driver needed.

2. Added additional python environment libraries for Waveshare devices, this runs as an additional installation step in the script.

3. Correctly markup the `/boot/firmware/config.txt` file for Waveshare devices, also ensuring that running the installation is idempotent and does not add multiple `dtoverlay` options.

4. Store the display type in `device.json` - this is used for instantiating the correct display.

6.  Maintain backwards compatibility with Inky operation for install.

8. Refactored the display framework so that `DisplayManager` creates a real instance of a concrete `AbstractDisplay` for either Inky or Waveshare devices.  The concrete instance is responsible for device specific operations.  
    - Created new display package for classes and _waveshare_epd_ drivers.
    - Waveshare has different drivers for different displays.  The parameter added to `device.json` is used to dynamically load the driver from the new package.
    - Refactor offers extensibility, to support a new screen manufacturer a new subclass of `AbstractDisplay` would be implemented with the device specific operations.
    - Included two drivers _epd7in3e.py_ and _epd7in3f.py_, to add additional Waveshare epd devices just drop the epd_X_in_Y.py file in the _waveshare_epd_ library.
    
9. Updated  documentation:
    - README.md with augmented installation instructions, Waveshare device extensibility.
    - troubleshooting.md with hints for troubleshooting Waveshare installations.
    - attribution.md with license information for Waveshare (MIT style).

PR #739, opened 2026-08-10
Accept a driver path for the -W install option

fabiomdiniz:
## Problem

`fetch_waveshare_driver` always builds the download URL from the common driver
directory:

```
https://raw.githubusercontent.com/waveshareteam/e-Paper/master/RaspberryPi_JetsonNano/python/lib/waveshare_epd/$WS_TYPE.py
```

Some models are not published there. The 3.6inch e-Paper (E), for example, only
ships under `E-paper_Separate_Program`, so `-W epd3in6e` fails the download and
the install aborts with "Ensure the model name is correct".

## Change

`-W` now also accepts the path to the driver within the waveshareteam/e-Paper
repository:

```bash
sudo bash install/install.sh -W E-paper_Separate_Program/3.6inch_e-Paper_E/RaspberryPi_JetsonNano/python/lib/waveshare_epd/epd3in6e.py
```

A new `resolve_waveshare_type()` splits that into the directory to fetch from
and the bare module name, so `WS_TYPE` ends up as `epd3in6e` — which is what
`update_config` writes to `device.json` and what `WaveshareDisplay` imports at
runtime. A leading `/` and a missing `.py` are both tolerated. A bare model
name takes the default directory and behaves exactly as before.

`epdconfig.py` is now fetched from the same directory as the driver, since the
separate programs ship their own copy, falling back to the common one if that
directory has none.

Also: a failed download no longer leaves the empty file behind that
`curl --fail -o` creates. Otherwise the `[ -f "$EPD_CONFIG_FILE" ]` check on the
next run reports it as "already exists" and silently skips the download.

## Testing

`bash -n install/install.sh` passes, and `resolve_waveshare_type` was exercised
directly:

| `-W` argument | `WS_TYPE` | driver URL |
|---|---|---|
| `epd7in3f` | `epd7in3f` | common dir (unchanged) |
| `epd3in6e.py` | `epd3in6e` | common dir |
| `E-paper_Separate_Program/.../waveshare_epd/epd3in6e.py` | `epd3in6e` | separate-program dir |
| leading `/`, no `.py` | `epd3in6e` | separate-program dir |

The URLs built for `epd3in6e.py`, its sibling `epdconfig.py`, and the unchanged
`epd7in3f.py` path all return 200. The installer has not been run end to end on
a Pi.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,11 @@ To install InkyPi, follow these steps:
     sudo bash install/install.sh -W epd7in3f
     ```
 
+    Some models are not published in the [common driver directory](https://github.com/waveshareteam/e-Paper/tree/master/RaspberryPi_JetsonNano/python/lib/waveshare_epd) and only ship under `E-paper_Separate_Program`. For those, pass the path to the driver within the [waveshareteam/e-Paper](https://github.com/waveshareteam/e-Paper) repository instead of the bare model name:
+    ```bash
+    sudo bash install/install.sh -W E-paper_Separate_Program/3.6inch_e-Paper_E/RaspberryPi_JetsonNano/python/lib/waveshare_epd/epd3in6e.py
+    ```
+
 
 After the installation is complete, the script will prompt you to reboot your Raspberry Pi. Once rebooted, the display will update to show the InkyPi splash screen.
 
--- a/install/install.sh
+++ b/install/install.sh
@@ -9,6 +9,11 @@
 #        -W <waveshare_device> (optional) Install for a Waveshare device,
 #                               specifying the device model type, e.g. epd7in3e.
 #
+#                               Alternatively the path to the driver within the
+#                               waveshareteam/e-Paper repository can be given,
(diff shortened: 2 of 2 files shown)


fabiomdiniz:
@acscpt this touches `fetch_waveshare_driver` and the `-W` handling you added in #103, so you're probably the best person to sanity-check it — particularly whether pulling `epdconfig.py` from the separate-program directory rather than the common one is right for those models.

I can't add a formal review request as an outside contributor. @fatihak, happy to adjust the approach if you'd rather this were an explicit second flag (e.g. `-P <path>`) than overloading `-W`.

phattmatt (approved):
Hi @fabiomdiniz,

Thanks for your PR!

I don't own any Waveshare displays to test completely, but running your improved `install.sh` seems to work as described.

The files are downloaded successfully and placed in the correct location.

The previous behaviour also seems to be working as before (no regressions).

Thanks for your contribution!

Matt.

PR #741, opened 2026-08-16
Feature: web preview

rusudorin:
Introduces a new button where one can preview the plugin before updating the screen.


rusudorin:
Note: I'm playing with github copilot with features I would find interesting. Wanted to share with you also if this is something you consider.

PR #742, opened 2026-08-16
perf: Skip no-op image enhancements

rusudorin:
ImageEnhance.enhance(1.0) still blends over the entire image, so the default settings ran four full-frame passes on every render. Skip any enhancement whose factor is 1.0. Output is identical since a factor of 1.0 returns the original pixels.


phattmatt (approved):
Hi @rusudorin,

Thanks for your PR, the changes look good!

Matt.

